PR #58, opened 2024-03-04
Use default tag 'latest' for modelkits that are missing one

amisevsk:
### Description

Copy OCI container semantics and use `latest` for the default tag when it's missing from an image reference.


amisevsk:
Closing in favor of https://github.com/jozu-ai/kitops/pull/143 (the same PR but from my fork rather than the main repository).

PR #61, opened 2024-03-04
Add Dockerfile for installing kit as a KServe ClusterStorageContainer

amisevsk:
### Description
Add a Dockerfile to enable using kit as a ClusterStorageContainer for KServe via the following CR:
```yaml
apiVersion: "serving.kserve.io/v1alpha1"
kind: ClusterStorageContainer
metadata:
  name: kitops
spec:
  container:
    name: kit-storage-initializer
    image: <kit-kserve-image>
    imagePullPolicy: Always
    env:
      - name: KIT_EXPORT_FLAGS
        value: ""       # Optional: Add extra flags for the `kit export` command
    resources:
      requests:
        memory: 100Mi
        cpu: 100m
      limits:
        memory: 1Gi     # Note: Default value, not measured
    supportedUriFormats:
    - prefix: kit://
```

This will allow downloading modelkits by specifying storageUris with `kit://` -- for example
```yaml
apiVersion: "serving.kserve.io/v1beta1"
kind: "InferenceService"
metadata:
  name: "torchserve"
spec:
  predictor:
    model:
      modelFormat:
        name: pytorch
      storageUri: kit://<modelkit-reference>
```

Note that KServe inference services generally assume a certain file structure for the model/configuration; as a result building dedicated modelkits for KServe is still likely required.

### DIFF
--- a/build/dockerfiles/KServe/README.md
+++ b/build/dockerfiles/KServe/README.md
@@ -1,0 +1,56 @@
+# Kitops ClusterStorageContainer image for KServe
+
+The Dockerfile in this directory is used to build an image that can run as a ClusterStorageContainer for [KServe](https://kserve.github.io/website/master/).
+
+## Building
+To build the image, `docker` is required. From the root of this repository, set the `$KIT_KSERVE_IMAGE` to the image tag you want to build and run
+```bash
+docker build . \
+  -f build/dockerfiles/KServe/kserve.Dockerfile \
+  -t $KIT_KSERVE_IMAGE \
+  --build-arg version="next" \
+  --build-arg gitCommit=$(git rev-parse HEAD)
+```
+
+## Installing
+The following process will create a new ClusterStorageContainer that uses `kit` to support KServe InferenceServices with storage URIs that have the `kit://` prefix.
+
+Create the following ClusterStorageContainer custom resource in a Kubernetes cluster with KServe installed
+```yaml
+apiVersion: "serving.kserve.io/v1alpha1"
+kind: ClusterStorageContainer
+metadata:
+  name: kitops
+spec:
+  container:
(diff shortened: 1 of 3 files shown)


build/dockerfiles/KServe/README.md
```diff
@@ -0,0 +1,56 @@
+# Kitops ClusterStorageContainer image for KServe
+
+The Dockerfile in this directory is used to build an image that can run as a ClusterStorageContainer for [KServe](https://kserve.github.io/website/master/).
+
+## Building
+To build the image, `docker` or `podman` is required. From the root of this repository, run
+```bash
```
gorkem, line 7: Let's use docker as part of the example. Podman may confuse people. I have already checked with docker, command works as expected

build/dockerfiles/KServe/README.md
```diff
@@ -0,0 +1,56 @@
...
+podman build . \
+  -f build/dockerfiles/KServe/kserve.Dockerfile \
+  -t $KIT_KSERVE_IMAGE \
```
gorkem, line 10: We should mention that this value should be set or substituted

gorkem (changes requested):

build/dockerfiles/KServe/README.md
amisevsk: Done; ideally this step isn't important since we'd eventually be building this image and publishing it on ghcr.io, which would move build to the bottom of this doc.

build/dockerfiles/KServe/README.md
amisevsk: Mentioned above -- how do we usually document a docker build like this 🤔

gorkem (approved):

PR #83, opened 2024-03-08
Add commands `kit inspect` and `kit info`

amisevsk:
### Description
Add commands to the CLI:

* `kit inspect [flags] MODELKIT` - Print the manifest for a ModelKit
* `kit info [flags] MODELKIT` - Print the config (Kitfile) for a ModelKit

Both commands will, by default, show a local ModelKit if it exists before checking remote repositories; each command supports the flag `-r|--remote` to force checking a remote in case it exists both locally and in the registry (e.g. in the case of a `latest` ModelKit that maybe out of date).

In testing `kit info`, I realized that the yaml library we're using (`sigs.k8s.io/yaml`) prints struct fields in alphabetic order, with no way to control this behavior. This meant that Kitfiles would be rendered as
```yaml
code: []
datasets: []
manifestVersion: v1.0.0
model: {}
package:
  license: <license>
  name: <name>
```

As a result, I've switched the repo back to using `gopkg.in/yaml.v3` (which preserves the expected order like JSON does) and reordered struct fields for better legibility.

### DIFF
--- a/pkg/cmd/info/cmd.go
+++ b/pkg/cmd/info/cmd.go
@@ -13,7 +13,6 @@ import (
 	"github.com/spf13/cobra"
 	"oras.land/oras-go/v2/errdef"
 	"oras.land/oras-go/v2/registry"
-	"sigs.k8s.io/yaml"
 )
 
 const (
@@ -69,7 +68,7 @@ func runCommand(opts *infoOptions) func(*cobra.Command, []string) {
 			}
 			output.Fatalf("Error resolving ModelKit: %s", err)
 		}
-		yamlBytes, err := yaml.Marshal(config)
+		yamlBytes, err := config.MarshalToYAML()
 		if err != nil {
 			output.Fatalf("Error formatting manifest: %w", err)
 		}
--- a/pkg/cmd/info/cmd.go
+++ b/pkg/cmd/info/cmd.go
@@ -16,19 +16,19 @@ import (
 )
 
 const (
-	shortDesc = `Show the configuration for a ModelKit`
-	longDesc  = `Print the contents of a ModelKit config to the screen.
+	shortDesc = `Show the configuration for a modelkit`
+	longDesc  = `Print the contents of a modelkit config to the screen.
 
-By default, kit will check local storage for the specified ModelKit. To see
-the configuration for a ModelKit stored on a remote registry, use the
+By default, kit will check local storage for the specified modelkit. To see
+the configuration for a modelkit stored on a remote registry, use the
 --remote flag.`
-	example = `# See configuration for a local ModelKit:
+	example = `# See configuration for a local modelkit:
 kit info mymodel:mytag
 
-# See configuration for a local ModelKit by digest:
+# See configuration for a local modelkit by digest:
 kit info mymodel@sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a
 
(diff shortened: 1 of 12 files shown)


pkg/cmd/info/cmd.go
```diff
@@ -0,0 +1,100 @@
...
+	"kitops/pkg/output"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"oras.land/oras-go/v2/errdef"
+	"oras.land/oras-go/v2/registry"
+)
+
+const (
+	shortDesc = `Show the configuration for a ModelKit`
```
gorkem, line 19: On the other commands we have used ModelKit as modelkit. We should either change this or the other rest of the docs

pkg/cmd/info/info.go
```diff
@@ -0,0 +1,63 @@
...
+		// Local only check
+		return getLocalConfig(ctx, opts)
+	}
+	if opts.checkRemote {
+		// Remote only check
+		return getRemoteConfig(ctx, opts)
+	}
+
+	// Check locally first; if not found check remote
+	manifest, err := getLocalConfig(ctx, opts)
```
gorkem, line 26: Should we actually check if the remote modelkit SHA matches the local one? This is for the case where somebody else moves the tag and now the same tag is referring to different versions on local and remote.

pkg/cmd/inspect/inspect.go
```diff
@@ -0,0 +1,55 @@
...
+		// Local only check
+		return getLocalManifest(ctx, opts)
+	}
+	if opts.checkRemote {
+		// Remote only check
+		return getRemoteManifest(ctx, opts)
+	}
+
+	// Check locally first; if not found check remote
+	manifest, err := getLocalManifest(ctx, opts)
```
gorkem, line 26: same comment as info. What happens if the local and remote modelkits that refer to the same tag are not a match?

gorkem (changes requested):

pkg/cmd/info/cmd.go
amisevsk: Yeah, I wasn't clear what our term was, since the readme is using `ModelKit`.

pkg/cmd/info/info.go
amisevsk: It's tricky -- we don't necessarily want to be pinging the remote registry if someone intends to work locally, and we can't necessarily show useful information in case the digest associated with a tag has diverged (e.g. did I build a new ModelKit locally or did the remote one update?). We can maybe improve this via timestamps or similar, though.

For now I'm opting for the semantics of docker/podman, where local vs remote tags can diverge and it's on the user to figure out what to do. I think docker goes one step further and only supports inspecting local images -- this may be a better choice for us as well. 

What do you think about simplifying `info` and `inspect` to work local-only by default, with the optional `-r|--remote` flag available to check remotes explicitly?

gorkem: It is a good idea to make remote checking explicit with `-r` flag. We can even print a message when we notice that remote and local are not pointing to the same SHA.

amisevsk: Done :+1:

pkg/cmd/info/cmd.go
amisevsk: Switched to `modelkit` for now, to match the other commands.

gorkem (approved):
lgtm

PR #89, opened 2024-03-11
Update how default config directory is handled

amisevsk:
### Description

Update how we resolve a default configuration directory for the CLI, in order of precedence:

* The value of the `--config` flag, if specified
* The value of the `$KITOPS_HOME` environment variable, if set
* A default OS-dependent value:
  * linux: `$XDG_DATA_HOME/kitops`, falling back to `~/.local/share/kitops`
  * windows: `%LOCALAPPDATA\kitops`
  * darwin: `~/Library/Caches/kitops`

If a suitable directory cannot be found (e.g. `$HOME` is not set), print an error message telling the user to either use the `--config` flag or set the `$KITOPS_HOME` environment variable.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -5,10 +5,7 @@ package cmd
 
 import (
 	"context"
-	"fmt"
 	"os"
-	"os/user"
-	"path/filepath"
 
 	"kitops/pkg/cmd/list"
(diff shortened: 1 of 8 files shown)


gorkem (approved):
LGTM

PR #92, opened 2024-03-11
Fix the `package.json` path in the github pages workflow

javisperez:
The package.json file lives inside `/docs`, not `/`.


bmicklea (approved):
LGTM

PR #101, opened 2024-03-12
New getting started doc

bmicklea:
* add getting-started.md doc
* add structure.md doc for Kitfile
* update readme.md to point to getting started


docs/src/docs/kitfile/structure.md
gorkem, line 1: Feels like this should be part of the Overview

bmicklea:
Ok, I think these are all the urgent docs tweaks we need for launch.

gorkem (approved):

PR #109, opened 2024-03-13
Fix the docs build

javisperez:
This PR fixes the docs build that is currently not passing.

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -134,8 +134,5 @@ export default defineConfig({
     }
   },
 
-  ignoreDeadLinks: [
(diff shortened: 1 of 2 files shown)


docs/.vitepress/config.mts
```diff
@@ -134,8 +134,5 @@ export default defineConfig({
     }
   },
 
-  ignoreDeadLinks: [
-    './CODE-OF-CONDUCT',
-    './GOVERNANCE'
-  ]
+  ignoreDeadLinks: true
```
javisperez, line 137: For now i've decided to ignore all dead links, because we're still kind of WIP with these.

gorkem (approved):

PR #122, opened 2024-03-14
Updated nightly build to generate signed binaries

nida-hasan:


gorkem (approved):

.github/workflows/nightly-build.yaml
amisevsk, line 1: I think signing will still need to run on a `macos-latest` runner ([doc](https://github.com/Bearer/gon?tab=readme-ov-file#usage))

> gon must be run on a macOS machine with XCode 11.0 or later. Code signing, notarization, and packaging all require tools that are only available on macOS machines.

PR #123, opened 2024-03-14
signed binaries for MacOS

nida-hasan:


.github/workflows/nightly-build.yaml
```diff
@@ -10,7 +10,7 @@ permissions:
   contents: write
 jobs:
   nightly-build:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
```
amisevsk, line 13: Not to rain on any parades but it looks like this dropped darwin-arm64: https://github.com/jozu-ai/kitops/releases/tag/next

nida-hasan: good catch! thanks!

amisevsk (approved):

PR #143, opened 2024-03-19
Use default tag 'latest' for modelkits that are missing one

amisevsk:
### Description

Copy OCI container semantics and use `latest` for the default tag when it's missing from an image reference.

Note: PR copied from https://github.com/jozu-ai/kitops/pull/58 to move head branch to my fork


amisevsk:
Closing to clear up clutter; we can reopen this PR if we decide to go down this path later.

PR #151, opened 2024-03-21
Return clearer error messages when failing to parse a reference

amisevsk:
### Description

Re-implement ParseReference to avoid unclear error messages. Instead, return a user-friendly error if e.g. the repository has an uppercase letter:

```
❯ kit tag mymodel:latest MyModel:latest
Failed to parse argument: failed to parse reference 'MyModel:latest': repository (MyModel) name must be lowercase
```


gorkem (approved):
LGTM 
The message`Failed to parse argument: failed to parse reference '1llamaTWO': repository (1llamaTWO) name must be lowercase` is a bit repetive. Should we change the first level of message to `Invalid argument`

PR #153, opened 2024-03-21
Add progress bars to long-running operations

amisevsk:
Add progress bars for kit push, pull, and unpack:

* Unpack
    unpack

* Pull (Push is the same)
    pull

Progress bars can be disabled via `--progress=false` and are disabled when the output is not a TTY. For push/pull, one progress bar is shown per blob downloaded; for unpack a progress bar is shown for the current step, below the regular output, which allows for regular logging (e.g. debug logs) without interfering with the bar.


gorkem:
Probably needs a test update

gorkem (approved):
LGTM... 
Feels like `ProgressLogger` may need another iteration for easier usage but I could not really imagine how we would simplfy it yet.

amisevsk:
Yeah, agreed. I'm not thrilled with it.

Issue #85, opened 2024-03-08
Allow referencing remote assets


bmicklea:
**Describe the problem your feature would solve**
When using very large datasets (e.g., image and video data especially), it's painful to have to clone them to my local machine in order to package up a ModelKit.

**Describe the solution you'd like**
It would be nice if you could provide a URI to the dataset in the Kitfile. I understand that that would trade download time for me for wait time as the ModelKit fetched the data in order to be built...as long as it doesn't happen on my machine 😄 

**Describe alternatives you've considered**
There isn't really an alternative beyond cloning everything locally.

**Additional context**
None

bmicklea:
**Decision:** We'll do this by referencing other ModelKits.

PR #156, opened 2024-03-23
added next-steps.md

bmicklea:
Added a Next Steps with Kit doc to cover authoring of a Kitfile and go into the Kit CLI commands that weren't covered in the Quick Start.

Updated the Quick Start to refer to the Next Steps.


docs/src/docs/next-steps.md
```diff
@@ -0,0 +1,232 @@
...
+* Make your own Kitfile
+* The power of `unpack`
+* Read the Kitfile or manifest from a ModelKit
+* Tag ModelKits and keep your registry tidy
+
+## Making your own Kitfile
+
+A Kitfile is the configuration document for your ModelKit. It's written in YAML so it's easy to read. There are four main parts to a Kitfile:
+1. ModelKit metadata in the `package` section, including the author, description, and license
+1. Path to the Jupyter notebook folder in the `code` section
```
gorkem, line 13: Perhaps we should say something like `Path to code assets, such as Jupyter notebooks` because code is not limited to Jupyter notebooks and can be either a file or folder.

docs/src/docs/next-steps.md
```diff
@@ -0,0 +1,232 @@
...
+1. Path to the datasets in the `datasets` section (you can have multiple datasets in the same page)
```
gorkem, line 14: `multiple datasets in the same page` -> `multiple datasets in the same ModelKit`

docs/src/docs/next-steps.md
```diff
@@ -0,0 +1,232 @@
...
+
+datasets:
+- name: training data
+  path: ./data/train.csv
+- description: validation data
+  name: validation data
+  path: ./data/validate.csv
+```
+
+Right now you can only build ModelKits from files on your local system...but don't worry we're already working towards allowing you to [reference remote files](https://github.com/jozu-ai/kitops/issues/85). For example, building a ModelKit from a local notebook and model, but a dataset hosted on DvC, S3, or anywhere else.
```
gorkem, line 66: I think we should remove this section. #85 will implement references to other ModelKits and this section is misleading.

docs/src/docs/next-steps.md
```diff
@@ -0,0 +1,232 @@
...
+
+Once you've authored a Kitfile for your AI/ML project you can pack it up and store it in your local or remote repository.
+
+```sh
+kit pack . -t film-slm:champion
+kit push docker.io/jozubrad/film-slm:champion
```
gorkem, line 72: Needs to be tagged with the registry reference before push
`kit pack . -t docker.io/jozubrad/film-slm:champion`

gorkem (changes requested):

docs/src/docs/next-steps.md
bmicklea: done

docs/src/docs/next-steps.md
bmicklea: done

docs/src/docs/next-steps.md
bmicklea: done

gorkem (approved):

PR #182, opened 2024-04-02
Update the outline format for the commands reference

javisperez:
This PR updates the outline for the command references docuemntation to not highlight subtitles for each command in the ToC (the one on the right).

## Preview

image


docs/src/docs/cli/cli-reference.md
javisperez, line 1: This whole file is auto generated by the CLI reference script, so i didnt change any of this. The real change is in the `cli-reference.header` only.

PR #197, opened 2024-04-08
Fix logos marquees on mobile

### LINKED ISSUE #193: Logo's Marquee issue in iOS Safari
**Describe the bug**
In iOS Safari, the Marquee of logos in the kitops site has a weird timing and spacing issue.

**To Reproduce**
1. In your mobile, using iOS Safari, go to https://kitops.ml
2. Scroll to the "What’s supported?" section
3. Wait a few seconds and you should notice the logos acting weird.

javisperez:
This PR fixes #193 by using a static width / height value on each logo instead of relying on dynamic flexbox dimensions.

It also switches from VueMarqueeSlider to Vue3Marquee as seems to be more maintained and stable.

### DIFF
--- a/docs/.vitepress/theme/assets/css/home.css
+++ b/docs/.vitepress/theme/assets/css/home.css
@@ -17,6 +17,6 @@
 }
 
 .marquee-gradients {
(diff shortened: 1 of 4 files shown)


gorkem (approved):
lgtm

PR #205, opened 2024-04-11
Add dev command

gorkem:


pkg/lib/harness/llm-harness.go
```diff
@@ -0,0 +1,156 @@
...
+
+	if _, err := os.Stat(pidFile); !os.IsNotExist(err) {
+		// Attempt to read the PID from the file.
+		pid, err := readPIDFromFile(pidFile)
+		if err != nil {
+			return fmt.Errorf("failed to read PID file: %w", err)
+		}
+		// Check if the process is still running.
+		if isProcessRunning(pid) {
+			return fmt.Errorf("A server process with PID %d is already running", pid)
```
amisevsk, line 51: Is there any way to instead return information about the running server? E.g.

```
Development server is already running at http://127.0.0.1:51021
Use "kit dev --stop" to stop the development server
```

amisevsk (commented):
Generally looks good, and works generally (except I get nonsense out of modelkits I've tested) 

There's a few places where we're using `fmt` instead of `output` for logging/errors that we should fix up.

Additionally I think running `kit dev` twice should just show that the server is already running instead of failing to start a new one.

pkg/lib/harness/llm-harness.go
gorkem: We could save the URL to somewhere I suppose.

frontend/dev-mode/src/components/ChatResults.vue
```diff
@@ -0,0 +1,140 @@
...
+          <input type="file" id="fileInput" class="opacity-0 absolute w-px h-px" @change.prevent="uploadImage">
+        </label>
+      </div>
+    </form>
+
+    <div class="mt-4 flex items-center justify-between">
+      <button
+        class="flex items-center gap-2 font-bold hover:text-cornflower"
+        @click="emit('leave')">
+        CHANGE PARAMETER
```
amisevsk, line 130: This line looks a little weird on the page, should maybe be "Change Parameters"?

amisevsk (approved):

amisevsk (commented):
On the completion page, every time the output wraps to a new line the page scrolls up a little which makes it hard to scroll with the completion on long output.

PR #212, opened 2024-04-12
Add support for .kitignore, fix paths handling when packing/unpacking

amisevsk:
### Description

Add support for `.kitignore` files. These files use the same semantics as `.dockerignore`:
* Each line is a path, files are ignored if the path is a prefix for their path
* Wildcards are supported: `*` - multiple characters, `?` - single character
* The `**` wildcard can be used to match any number of directories (e.g. `**/*.md` will ignore all `.md` files no matter where they are
* The prefix `!` can be used to negate matching, to reinclude a pattern that is excluded previously (e.g. adding `!README.md`) will include the readme even if it's ignored via `**/*.md`.

In addition, this PR fixes a few issues with how we were handling paths:

* Using a non `.` path for context dir (in pack) or target dir (in unpack) now works without strange behaviour with some relative paths (if something resolved to `.` it would be placed in the working directory instead of the target directory).
* It's now possible to use `.` as a path in Kitfiles, allowing e.g. a simple Kitfile where the model lives in the same directory as the Kitfile (previously, everything was assumed to be a subdirectory)
* Kit can now handle intersecting paths without compressing files twice, e.g. in the Kitfile below, the `my-model` model will not include the subdirectory `main-dir/intersect-dir` as it is included in the `my-dataset` dataset
    ```yaml
    manifestVersion: 1.0.0
    package:
      name: test-intersecting
    model:
      name: my-model
      path: main-dir
    datasets:
      - name: my-dataset
        path: main-dir/intersect-dir
    ```

The above also mean it's possible to write Kitfiles like
```yaml
manifestVersion: 1.0.0
package:
  name: test-intersecting
model:
  name: my-model
  path: ./my-model
code:
  - path: .
```

There's more detail on changes in commit messages.

### Linked issues
Closes https://github.com/jozu-ai/kitops/issues/119
Required for https://github.com/jozu-ai/kitops/issues/85

### Testing
To come...


pkg/lib/filesystem/ignore.go
```diff
@@ -0,0 +1,137 @@
...
+	ignoreFileMatches, err := pm.ignoreFileMatcher.MatchesOrParentMatches(path)
+	if err != nil {
+		return false, err
+	}
+	if ignoreFileMatches {
+		return true, nil
+	}
+	// ignore file doesn't exclude the current path, check if it should be excluded
+	// since it's included in another layer
+	for _, layer := range pm.layers {
```
gorkem, line 71: Feels like there should be a guard somewhere here to skip the currently packaged layer.

gorkem (changes requested):
When I test this with` ghcr.io/jozu-ai/modelkit-examples/tensorflow-image:latest` It fails to package files under the directories.

amisevsk:
> When I test this with ghcr.io/jozu-ai/modelkit-examples/tensorflow-image:latest It fails to package files under the directories.

Good catch, I wasn't cleaning paths before checking if they match, which led to `./data` not matching `data`.

pkg/lib/filesystem/ignore.go
amisevsk: How do you mean? This check is a little weird but I think it hits the main issue -- given a Kitfile
```yaml
manifestVersion: 1.0.0
package:
  name: test-intersecting
model:
  name: test-main
  path: main-dir
datasets:
  - name: test-intersect
    path: main-dir/intersect-dir
```
there are two cases:

* while processing the model (`main-dir` path), we want to ignore all paths prefixed with `main-dir/intersect-dir`
* while processing the dataset (`main-dir/intersect-dir` path), we _don't_ want to ignore all paths prefixed `main-dir` since that would ignore the whole layer

Worth noting that this was broken until the latest commit if you used `./main-dir` in the Kitfile, since the clean version of that path is `main-dir`, which breaks the prefix check.

gorkem: yes. I was using `./main-dir`

gorkem (approved):

Issue #196, opened 2024-04-08
Scratch builds for dev command from MacOS from branch


gorkem:

gorkem:
Closing as a scratch build was done from my fork and feature branch is about to be merged

PR #214, opened 2024-04-15
split platform builds

gorkem:
Splits the release builds to corresponding platforms. Introduces go-releaser configuration for each platform. adds a new workflow that uses GH CLI to do the publish. 

related to #196


amisevsk (approved):
LGTM, sorry I missed this for a couple days.

Minor request: some files have no trailing newline, others have multiple

.github/workflows/platform-release.yaml
```diff
@@ -0,0 +1,182 @@
...
+  release:
+    runs-on: ubuntu-latest
+    needs: [build-linux, build-macos, build-windows]
+    steps:
+      - name: Merge built artifacts
+        uses: actions/download-artifact@v4
+        with:
+          path: dist
+          pattern: dist-*
+          merge-multiple: true
```
amisevsk, line 132: I'm not sure how artifacts work -- if we run two releases within a week of each other will they see both artifacts? I suppose the second build would overwrite them?

gorkem: You can only download artifacts that were uploaded during the same workflow run or use the UI to navigate to the workflow run and download. I have also done a shortened retention, the default is 90 days.

amisevsk: That makes sense :+1:

gorkem:
> LGTM, sorry I missed this for a couple days.
> 
> Minor request: some files have no trailing newline, others have multiple

fixed

PR #232, opened 2024-04-17
Add podcast blogpost

javisperez:
For real now... 

This PR adds the blog posts to the proper json file and adds `KitOps` and `MLOps` to the custom-capitalization dictionary.

## Preview

image


Jwilliamsr (commented):
Thanks for adding podcast tag. 

Can you remove the two copied tags from the long description?

PR #238, opened 2024-04-18
update next steps

### LINKED ISSUE #237: Fix minor bugs in documentation
**Describe the bug**
There are a few places in the next-steps documentation where details are missing that can lead users to errors.

**To Reproduce**
For example, we are not clear on the necessity of the dash at the start of each dataset or codebase in the Kitfile. We are also unclear about the units that `size` uses in the manifest. There were minor naming inconsistencies in `unpack` versus `pull` and `pack` versus `push`

**Version**
latest

bmicklea:
### Description
* Updated next steps with more details, and to remove "gotchas"
* Updated the ModelKit spec.md to clarify that `size` is measured in KB

### Linked issues
fixes #237

### DIFF
--- a/docs/src/docs/next-steps.md
+++ b/docs/src/docs/next-steps.md
@@ -9,23 +9,49 @@ In this guide you'll learn how to:
 ## Making your own Kitfile
 
 A Kitfile is the configuration document for your ModelKit. It's written in YAML so it's easy to read. There are four main parts to a Kitfile:
+
 1. The `package` section: Metadata about the ModelKit, including the author, description, and license
-1. The `code` section: Path to any codebases related to the project, including Jupyter notebook folders
-1. The `datasets` section: Path to any datasets
-1. The `model` section: Path to the serialized model
+1. The `code` section: Path and information about codebases related to the project, including Jupyter notebook folders
+1. The `datasets` section: Path and infromation on included datasets
+1. The `model` section: Path and information on the serialized model
 
 A Kitfile only needs the `package` section, plus one of the other sections.
 
-The `datasets` and `code` sections can include multiple assets, but the `model` section can only contain a single model (you can chain models by using multiple ModelKits).
+The `package` and `model` sections can only contain a single model (you can chain models by using multiple ModelKits).
 
-Here's an example Kitfile:
+The `datasets` and `code` sections can include multiple assets. Each asset is indicated by starting with a dash. The dash is required even if you are only packaging a single dataset or codebase.
+
+Here's a snippet of a KitFile that contains two datasets, notice that each starts with "-":
 
 ```yaml
-manifestVersion: v1.0.0
+datasets:
+- description: Forum postings from photo sites
+  name: training data
+  path: ./data/forum-to-2023-train.csv
+  
+- name: validation data
+  path: ./data/test.csv
+```
+
+...and here's an example of a single codebase, it still needs to start with a "-":
+
+```yaml
+code:
+- description: Jupyter notebook with model training code in Python
+  path: ./notebooks
+```
+
+Any relative paths defined within the Kitfile are interpreted as being relative to the context directory sent to the `kit pack` command.
+
+### Kitfile Examples
+
+Here's a complete Kitfile example, with a model, two datasets, and a codebase:
+
+```yaml
+manifestVersion: 1.0
 
 package:
-  authors:
-  - Jozu
+  authors: Jozu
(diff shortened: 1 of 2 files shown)


docs/src/docs/next-steps.md
```diff
@@ -9,23 +9,49 @@ In this guide you'll learn how to:
 ## Making your own Kitfile
 
 A Kitfile is the configuration document for your ModelKit. It's written in YAML so it's easy to read. There are four main parts to a Kitfile:
+
 1. The `package` section: Metadata about the ModelKit, including the author, description, and license
-1. The `code` section: Path to any codebases related to the project, including Jupyter notebook folders
-1. The `datasets` section: Path to any datasets
-1. The `model` section: Path to the serialized model
+1. The `code` section: Path and information about codebases related to the project, including Jupyter notebook folders
+1. The `datasets` section: Path and infromation on included datasets
```
amisevsk, line 15:
```suggestion
1. The `datasets` section: Path and information on included datasets
```

docs/src/docs/next-steps.md
```diff
@@ -9,23 +9,49 @@ In this guide you'll learn how to:
...
+
+Here's a complete Kitfile example, with a model, two datasets, and a codebase:
+
+```yaml
+manifestVersion: 1.0
 
 package:
-  authors:
-  - Jozu
+  authors: Jozu
```
amisevsk, line 54: Authors is also a yaml list, so the formats are
```yaml
authors:
  - Jozu
```
or
```yaml
authors: ["Jozu"]
```

docs/src/docs/next-steps.md
```diff
@@ -9,23 +9,49 @@ In this guide you'll learn how to:
...
+datasets:
+- description: Forum postings from photo sites
+  name: training data
+  path: ./data/forum-to-2023-train.csv
+  
+- name: validation data
+  path: ./data/test.csv
+```
+
+...and here's an example of a single codebase, it still needs to start with a "-":
```
amisevsk, line 36: I wonder if there is a way of addressing this that hits both audiences (those familiar with yaml and those who aren't). The reason the `-` is needed is just because this is a list of objects, similar to how the JSON equivalent would be

```json
"datasets": [
  {
    "description": "desc",
    "name": "training data",
    "path": "./data/one.csv"
  },
  {
    "name": "training data",
    "path": "data/two.csv"
  }
]
```

We should figure out a way to validate the Kitfile in CLI to surface these tips -- I've created https://github.com/jozu-ai/kitops/issues/239 for this

docs/src/docs/next-steps.md
```diff
@@ -51,29 +77,49 @@ code:
...
   path: ./data/validate.csv
 ```
 
-Once you've authored a Kitfile for your AI/ML project you can pack it up and store it in your local or remote repository.
+More information on Kitfiles can be found in the [Overview](./kitfile/kf-overview.md) and [Format](./kitfile/format.md) documentation.
 
-If you don't provide a registry Kit assumes you're operating against your local registry. For example:
+### Using the Kitfile to Pack a ModelKit
+
+When you're done writing the Kitfile, name it `kitfile` without an extension. Now you can use the `kit pack` command to build your ModelKit.
```
amisevsk, line 101: I think our current "default" name is `Kitfile` with a capital K, though we can change this. If you unpack, you'll get a `Kitfile` instead of a `kitfile`.

docs/src/docs/next-steps.md
```diff
@@ -179,13 +225,15 @@ kit inspect mymodel:challenger
 }
 ```
 
-## Tag ModelKits and keep your registry tidy
+`size` is shown in kilobytes (KB). For more information on the manifest can be found in the [specification documentation](./modelkit/spec.md).
```
amisevsk, line 228: It's actually raw bytes (divide by 1,000 for KB, 1,000,000 for MB)

docs/src/docs/next-steps.md
bmicklea: FWIW `kitfile` worked on my machine as well but I've updated it to say `Kitfile` in the latest commit to be consistent with the pulled result.

docs/src/docs/next-steps.md
bmicklea: fixed in latest commit

docs/src/docs/next-steps.md
bmicklea: Thanks, fixed in latest commit

gorkem (approved):
lgtm

PR #261, opened 2024-04-24
Avoid copying layer into local storage during pack

amisevsk:
### Description
Avoid requiring a data copy during the packing process to cut down on peak storage usage and slightly increase speed. Previously, the process was

1. Compress layer files to a `.tar.gz` in `$TMPDIR` and read digest from file
2. Construct a manifest and push the `.tar.gz` into local OCI storage (requires copying)
3. Remove the temporary `.tar.gz` from step 1.

Since OCI storage is content-addressible, instead of pushing into OCI storage during step 2, we instead copy the temporary `.tar.gz` directly into its expected location and verify that the OCI store has picked it up correctly.

### Linked issues
N/A

### DIFF
--- a/pkg/lib/repo/local.go
+++ b/pkg/lib/repo/local.go
@@ -35,6 +35,7 @@ import (
 type LocalStorage interface {
 	GetRepo() string
 	GetIndex() (*ocispec.Index, error)
+	getStorePath() string
 	oras.Target
 	content.Deleter
 	content.Untagger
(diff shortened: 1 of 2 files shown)


amisevsk:
Performance wise, it's not really much faster assuming the disk is fast -- 77s vs 81s for `ghcr.io/jozu-ai/llama-2:7b-chat-q4_0` on my machine, but it should avoid some storage issues we run into with larger models.

gorkem (approved):
lgtm

PR #267, opened 2024-04-25
Add validation for Kitfiles

### LINKED ISSUE #81: kit pack error
**Describe the bug**
When trying to run `kit pack` with the attached Kitfile I get an error:
```
Failed to pack model kit: error resolving /Users/bmicklea/Documents/GitHub: lstat /Users/bmicklea/Documents/GitHub/Users: no such file or directory
```

It looks like it's appending a second `Users` to the path, although that's weird because the path actually has more to it...

[kitfile.txt](https://github.com/jozu-ai/kitops/files/14530215/kitfile.txt)

**To Reproduce**
1. Run `kit pack ./` with the attached kitfile (it'll need to have the .txt extension removed)
2. See error

**Version**
Next

amisevsk:
### Description
Add some basic validation to Kitfiles, printing an error if

* Multiple layers share the same path (makes unpacking/ignoring unclear)
* Any layer is an absolute path rather than relative
* Model part's type field is invalid: (max 64 characters, alphanumeric with `.`, `-`, and `_`; must start with alphanumeric)

### Testing
You can see all of the failure cases by trying to pack the following Kitfile

<details>
<summary>Kitfile</summary>

```yaml
manifestVersion: 1.0.0
package:
  name: test-kf
model:
  name: my-model
  path: path1
  parts:
    - name: part1
      path: path1
    - name: part2
      path: /abs/path
    - name: part3
      path: valid-path
      type: test type with space
    - name: part4
      path: also-valid
      type: really-long-but-otherwise-valid-type---this-line-is-65-chars-long
datasets:
  - name: dataset1
    path: path1
  - name: dataset2
    path: path3
  - name: dataset3
    path: /abs/path/2
code:
  - path: path1
  - path: path3
  - path: /abs/path/3
```
</details>

### Linked issues

Closes #81

### DIFF
--- a/pkg/lib/kitfile/validate.go
+++ b/pkg/lib/kitfile/validate.go
@@ -1,0 +1,72 @@
+// Copyright 2024 The KitOps Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package kitfile
+
+import (
+	"fmt"
+	"path"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"kitops/pkg/artifact"
+)
+
+// ValidateKitfile returns an error if the parsed Kitfile is not valid. The error string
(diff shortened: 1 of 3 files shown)


nida-hasan (approved):

pkg/lib/kitfile/validate.go
```diff
@@ -0,0 +1,88 @@
...
+	"path"
+	"path/filepath"
+	"regexp"
+	"slices"
+	"strings"
+
+	"kitops/pkg/artifact"
+)
+
+const partTypeMaxLen = 64
```
gorkem, line 30: Why 64?

gorkem (approved):

pkg/lib/kitfile/validate.go
amisevsk: Arbitrary choice -- big enough that nobody should hit the limit but also not too big since we can increase this without breaking any Kitfiles.

PR #272, opened 2024-04-28
Improve error handling in dev command

gorkem:
Uses Errorf instead of Fatalf and fixes missing returns. ↩️

### DIFF
--- a/pkg/cmd/dev/cmd.go
+++ b/pkg/cmd/dev/cmd.go
@@ -22,6 +22,7 @@ import (
 	"kitops/pkg/lib/filesystem"
 	"kitops/pkg/output"
 	"net"
+	"os"
 	"runtime"
 	"strconv"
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/dev/cmd.go
```diff
@@ -105,19 +105,22 @@ func runCommand(opts *DevOptions) func(cmd *cobra.Command, args []string) {
...
+				return
 			}
 			output.Infoln("Development server stopped")
 			return
 		}
 		err := runDev(cmd.Context(), opts)
 		if err != nil {
-			output.Fatalf("Failed to start dev server: %s", err)
+			output.Errorf("Failed to start dev server: %s", err)
+			return
```
amisevsk, line 123: We might want to still use `os.Exit(1)` in this case instead of `return`. Returning will result in the command exiting as "successful".

amisevsk (approved):

PR #277, opened 2024-04-30
Add more info to output of `kit inspect`

### LINKED ISSUE #271: Improve inspect output
It should be possible to get additional information such as the Digest of the image. 

Other fields that can be considered are the cli version and creation date.

amisevsk:
### Description
Add additional information to the output of `kit inspect`:
```jsonc
{
  "digest": "<modelkit-digest>",
  "cliVersion": "<cli-version>",
  "kitfile": {
    // Kitfile
  },
  "manifest": {
    // Manifest  
  }
}
```

Note this change does mean that both `kit inspect` and `kit info` produce the Kitfile for the modelkit. I've left both commands intact because `kit info` outputs the Kitfile in YAML format (easier to read) whereas `inspect` outputs JSON. We can't yet produce yaml output for inspect since the OCI manifest struct does not have yaml tags and `gopkg.in/yaml.v3` does not respect json tags.

Note: to test version labelling, you will need to build `kit` with the version fields set:
```bash
go build -o kit -ldflags "-s -w -X kitops/pkg/cmd/version.Version=v0.0.1+test"
```

### Linked issues
Closes #271

### DIFF
--- a/pkg/cmd/inspect/inspect.go
+++ b/pkg/cmd/inspect/inspect.go
@@ -19,6 +19,7 @@ package inspect
 import (
 	"context"
 	"fmt"
+	"kitops/pkg/artifact"
 	"kitops/pkg/lib/constants"
 	"kitops/pkg/lib/repo"
 
@@ -30,6 +31,7 @@ import (
 type inspectInfo struct {
 	Digest     digest.Digest     `json:"digest,omitempty" yaml:"digest,omitempty"`
 	CLIVersion string            `json:"cliVersion,omitempty" yaml:"cliVersion,omitempty"`
+	Kitfile    *artifact.KitFile `json:"kitfile,omitempty" yaml:"kitfile,omitempty"`
 	Manifest   *ocispec.Manifest `json:"manifest,omitempty" yaml:"manifest,omitempty"`
 }
 
@@ -47,7 +49,7 @@ func getLocalManifest(ctx context.Context, opts *inspectOptions) (*inspectInfo, 
 	if err != nil {
 		return nil, fmt.Errorf("failed to read local storage: %w", err)
 	}
-	desc, manifest, err := repo.ResolveManifest(ctx, store, opts.modelRef.Reference)
+	desc, manifest, config, err := repo.ResolveManifestAndConfig(ctx, store, opts.modelRef.Reference)
 	if err != nil {
 		return nil, err
 	}
@@ -58,6 +60,7 @@ func getLocalManifest(ctx context.Context, opts *inspectOptions) (*inspectInfo, 
 	return &inspectInfo{
 		Digest:     desc.Digest,
 		CLIVersion: version,
+		Kitfile:    config,
 		Manifest:   manifest,
 	}, nil
 }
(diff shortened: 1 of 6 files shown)


pkg/cmd/inspect/inspect.go
```diff
@@ -19,30 +19,53 @@ package inspect
...
 
-func getLocalManifest(ctx context.Context, opts *inspectOptions) (*ocispec.Manifest, error) {
+func getLocalManifest(ctx context.Context, opts *inspectOptions) (*inspectInfo, error) {
 	storageRoot := constants.StoragePath(opts.configHome)
 	store, err := repo.NewLocalStore(storageRoot, opts.modelRef)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read local storage: %w", err)
 	}
-	return repo.ResolveManifest(ctx, store, opts.modelRef.Reference)
+	desc, manifest, config, err := repo.ResolveManifestAndConfig(ctx, store, opts.modelRef.Reference)
```
gorkem, line 52: Can we replace `config` with `kitfile`? I always have to think for a few secs what is config in this context.

pkg/cmd/inspect/inspect.go
```diff
@@ -19,30 +19,53 @@ package inspect
...
+	return &inspectInfo{
+		Digest:     desc.Digest,
+		CLIVersion: version,
+		Kitfile:    config,
+		Manifest:   manifest,
+	}, nil
 }
 
-func getRemoteManifest(ctx context.Context, opts *inspectOptions) (*ocispec.Manifest, error) {
+func getRemoteManifest(ctx context.Context, opts *inspectOptions) (*inspectInfo, error) {
```
gorkem, line 68: Is this repeating too much of the `getLocalManifest`? perhps we need to refactor this logic after repositroy creation to a function.

pkg/lib/repo/repo.go
```diff
@@ -217,26 +217,30 @@ func GetConfig(ctx context.Context, store content.Storage, configDesc ocispec.De
 }
 
 // ResolveManifest returns the manifest for a reference (tag), if present in the target store
-func ResolveManifest(ctx context.Context, store oras.Target, reference string) (*ocispec.Manifest, error) {
+func ResolveManifest(ctx context.Context, store oras.Target, reference string) (ocispec.Descriptor, *ocispec.Manifest, error) {
 	desc, err := store.Resolve(ctx, reference)
 	if err != nil {
-		return nil, fmt.Errorf("reference %s not found in remote repository: %w", reference, err)
+		return ocispec.DescriptorEmptyJSON, nil, fmt.Errorf("reference %s not found in remote repository: %w", reference, err)
```
gorkem, line 223: Should the error message change to remove "remote". I think this method can be called for local repo too.

gorkem (approved):
I have some suggestions but otherwise  :ok_hand:

pkg/cmd/inspect/inspect.go
amisevsk: It all stems from the [original sin](https://github.com/jozu-ai/kitops/blob/bdceda24a2994947bbb2438103a895df278e4171/pkg/lib/storage/local.go#L61) :D

amisevsk:
Applied fixes suggested above

PR #279, opened 2024-04-30
Include current CLI version in generated docs page

### LINKED ISSUE #146: version number for the cli reference
The version number that is generated for the documentation site should have a reference to the version of the CLI it is generated with.

amisevsk:
### Description
Add version to CLI reference page, now (slightly) less ugly:

Screenshot 2024-05-15 at 5 53 07 PM

### Linked issues
Closes #146

### DIFF
--- a/.github/workflows/platform-release.yaml
+++ b/.github/workflows/platform-release.yaml
@@ -201,6 +201,7 @@ jobs:
           git config --global user.name "${GITHUB_ACTOR}"
           git config --global user.email "${GITHUB_ACTOR_ID}+${GITHUB_ACTOR}@users.noreply.github.com"
 
+          (cd docs; npm pkg set version=$TAG_NAME)
           ./docs/src/docs/cli/generate.sh > /dev/null
           git add --all
           git commit -m "docs: update CLI documentation for ${{ github.ref_name }}"
--- a/docs/.vitepress/theme/components/VersionInfo.vue
+++ b/docs/.vitepress/theme/components/VersionInfo.vue
@@ -1,0 +1,15 @@
+<script setup lang="ts">
+import { version } from '../../../package.json'
+</script>
+
+<template>
+  <div class="flex justify-between items-center">
+    <h1 id="kit-cli-reference" tabindex="-1">
+      Kit CLI Reference
+      <a class="header-anchor" href="#kit-cli-reference" aria-label="Permalink to &quot;Kit CLI Reference&quot;">​</a>
+    </h1>
+    <div class="text-gray-05 text-med font-mono">
+      {{version}}
+    </div>
+  </div>
+</template>
--- a/docs/src/docs/cli/cli-reference.header
+++ b/docs/src/docs/cli/cli-reference.header
@@ -1,5 +1,8 @@
 ---
 outline: 2
+title: "Kit CLI Reference"
 ---
-# Kit CLI Reference
-
+<script setup>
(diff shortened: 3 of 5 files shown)


amisevsk:
Just realized that this will look strange with our current version format `0.2.0-93739be`

.github/workflows/platform-release.yaml
```diff
@@ -190,6 +190,7 @@ jobs:
           git config --global user.name "${GITHUB_ACTOR}"
           git config --global user.email "${GITHUB_ACTOR_ID}+${GITHUB_ACTOR}@users.noreply.github.com"
 
+          (cd docs; npm pkg set version=$TAG_NAME)
```
javisperez, line 193: As a frontend developer, i approve this message 🤝

docs/src/docs/cli/cli-reference.header
```diff
@@ -1,5 +1,8 @@
 ---
 outline: 2
+title: "Kit CLI Reference"
 ---
-# Kit CLI Reference
-
+<script setup>
+  import { version } from '../../../package.json'
+</script>
+# Kit CLI Reference {{ version }}
```
javisperez, line 8: We definitely will want to update this :D but later, ill leave this to Annika but i think that having the version in the h1 is not ideal as it changes with each release.

javisperez (approved):

javisperez:
> Just realized that this will look strange with our current version format `0.2.0-93739be`

As it lives in the package.json maybe we can just show it in the sidebar, in a smaller, less "eye catching" way? just throwing ideas...

docs/src/docs/cli/cli-reference.header
amisevsk: Yeah, I wasn't sure where to put it honestly.

gorkem (approved):

gorkem:
I agree, I was not expecting it on the header. Somewhere on the sidebar makes sense.

amisevsk:
We can put it wherever we like, I just don't see a good spot in the sidebar (unless we want to do a dropdown + previous versions, but that seems like a v1.0 problem).

bmicklea:
Can you drop it at the bottom of the file? It's something we might want to check for consistency but we're not doing a fancy drop-dow-to-view-old-versions thing.

annigro (approved):

gorkem (approved):

docs/.vitepress/theme/components/VersionInfo.vue
```diff
@@ -0,0 +1,15 @@
+<script setup lang="ts">
+import { version } from '../../../package.json'
+</script>
+
+<template>
+  <div class="flex justify-between items-center">
+    <h1 id="kit-cli-reference" tabindex="-1">
+      Kit CLI Reference
```
javisperez, line 8: All these changes are good to me, but in the future, and for a bigger reusability, we want this line to be a `<slot />` so we can pass any title content that we want, and therefor, reuse it from anyplace.

javisperez (approved):

docs/.vitepress/theme/components/VersionInfo.vue
amisevsk: I was going to do something like that, but I couldn't figure out a good way to parameterize the id/anchor bits -- if you've got tips I'm happy to fix it :+1:

TBH I copied the anchor from what Vite autogens for a markdown `#` 🙂

PR #287, opened 2024-05-01
Add command to print logs for dev server

amisevsk:
### Description
Add flag `--logs` to kit dev to allow printing the log file

### Linked issues
N/A

### DIFF
--- a/pkg/cmd/dev/cmd.go
+++ b/pkg/cmd/dev/cmd.go
@@ -16,10 +16,6 @@
 package dev
 
 import (
-	"context"
-	"fmt"
-	"kitops/pkg/lib/constants"
-	"kitops/pkg/lib/filesystem"
 	"kitops/pkg/lib/harness"
 	"kitops/pkg/output"
 	"net"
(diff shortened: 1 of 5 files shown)


gorkem (commented):
I think we were on the shaky ground with the `--stop` and with this one I think we might need to go into the subcommands to dev command. Alternately this could be a flag to tail the log. WDYT?

amisevsk:
Yeah, I noticed while writing it that we're effectively doing subcommands in the switch at this point. I can rewrite to be subcommands relatively easily.

amisevsk:
@gorkem I've reworked `dev` to have three subcommands:

* `kit dev start`: start server (previously: `kit dev`)
* `kit dev stop`: stop server (previously: `kit dev --stop`)
* `kit dev logs`: show logs (previously `kit dev --logs`)

pkg/cmd/dev/cmd.go
```diff
@@ -29,74 +26,58 @@ import (
...
-				return err
-			}
-			opts.port = availPort
-		}
+func DevCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "dev <directory> [flags]",
+		Short:   devShortDesc,
+		Long:    devLongDesc,
+		Example: devExample,
```
gorkem, line 34: Is example needed?

gorkem (approved):

pkg/cmd/dev/cmd.go
amisevsk: I included it as an additional hint, since running `kit dev` just prints the help page now (whereas before it would start up the server):
```
❯ kit dev
Run models locally (experimental)

Description:
  Start a local server and interact with a model in the browser

Usage:
  kit dev [command]

Examples:
  kit dev start

Available Commands:
  logs        View logs for development server
  start       Start development server (experimental)
  stop        Stop development server
```

gorkem: :shipit:

PR #288, opened 2024-05-01
Pass the `slot_id` parameter as a number

javisperez:
The `slot_id` parameter is wrongly sent as a string, should be a number.

This PR fixes that :).

This PR also adds some tiny UI fixes that i noticed.

### DIFF
--- a/frontend/dev-mode/src/components/ChatResults.vue
+++ b/frontend/dev-mode/src/components/ChatResults.vue
@@ -71,7 +71,7 @@ useResizeObserver(resultsContainer, () => {
 </script>
 
 <template>
-<div role="presentation" class="w-full pb-16 max-w-3xl mx-auto flex-1 flex flex-col">
+<div role="presentation" class="w-full max-w-3xl mx-auto flex-1 flex flex-col">
   <div class="flex-1 h-full flex flex-col justify-end" ref="resultsContainer">
     <template v-for="([actor, response], index) in session.transcript" :key="index">
       <div class="font-bold mt-6">{{ template(actor) }}</div>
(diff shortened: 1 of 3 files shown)


frontend/dev-mode/src/views/PromptsView.vue
```diff
@@ -63,7 +63,7 @@ const {
   runCompletion,
   stop,
   uploadImage
-} = useLlama(parameters.value)
+} = useLlama({ ...parameters.value, slot_id: -1 })
```
javisperez, line 66: this hard `-1` number is because the first time, it must be -1 but then should just use any value that came back from the api response.

frontend/dev-mode/src/views/PromptsView.vue
```diff
@@ -40,7 +39,8 @@ const DEFAULT_PARAMS_VALUES = {
   image_data: [],
   cache_prompt: true,
   api_key: '',
-  prop_order: '',
+  prop_order: undefined,
```
javisperez, line 42: if this is untouched we dont want to send an empty string.

frontend/dev-mode/src/composables/useLlama.ts
```diff
@@ -70,19 +84,7 @@ export const DEFAULT_SESSION: Session = {
...
-  isGenerating: Ref<boolean>;
-  isChatStarted: Ref<boolean>;
-  isPending: Ref<boolean>;
-  chat: (msg: string) => Promise<void>;
-  runCompletion: () => void;
-  stop: (e: Event) => void;
-  reset: (e: Event) => void;
-  uploadImage: (e: Event) => void;
-} {
+export default function useLlama(params?: UserParameters, localSession?: Session): LlamaComposableResponse {
```
javisperez, line 87: extracted these types to its own Type to keep this line cleaner.

frontend/dev-mode/src/components/ChatResults.vue
```diff
@@ -82,14 +82,17 @@ useResizeObserver(resultsContainer, () => {
 
       <CopyTextButton
         :text="joinResponse(response as TranscriptMessage[])"
-        :class="{ '!visible': ((response as TranscriptMessage[])[0]).id_slot !== undefined }"
+        :class="{
+          '!visible': ((response as TranscriptMessage[])[0]).id_slot !== undefined,
+          'hidden': template(actor) === session.user
```
javisperez, line 87: Just minor UI improvement.

gorkem (approved):

PR #298, opened 2024-05-06
Use no compression for model layers by default

### LINKED ISSUE #257: Improve pack and unpack speed
Modelkits with large files can take a long time to pack/unpack due to gzip being slow. We can speed this up but need to consider options carefully as changing the compression format will change model digests.

Choose another option for storage that is quicker
* No compression is fastest (obviously) but takes up the most space. This might not be a huge issue if the bulk of data in a modelkit isn't compressible anyways
* Gzip on the fastest compression option appears (at first glance) to be a lot faster with no real loss in compression ratio
* Zstd seems to be the best option in terms of size/speed, but the options for integrating it are limited:
    * [`github.com/klauspost/compress/zstd`](https://pkg.go.dev/github.com/klauspost/compress/zstd) is a Go implementation of the algorithm, but abstracts over compression levels (in my testing levels here do not correspond to the C implementation)
    * [`github.com/DataDog/zstd`](https://github.com/DataDog/zstd) is a wrapper on the C implementation, but requires CGO to build.

We could also make storage type configurable (gzip, no compression, zstd) (e.g. in the Kitfile). This would lead to a situation where the same modelkit data potentially packs into modelkits with different digests, though.

**Additional context**

amisevsk:
### Description
This PR is adapted from the [compression-opts](https://github.com/jozu-ai/kitops/tree/compression-opts) branch. See discussion on #257 for details. This PR drops the `zstd` option from that branch.

This PR adds the flag `--compression` to `kit pack`, with valid options `none`, `gzip`, and `gzip-fastest`:
* `none`: no compression of layers, store layers as plain tarballs
* `gzip`: current default behavior, compress layer tarballs with gzip
* `gzip-fastest`: compress layer tarballs with gzip at the lowest compression level

The new default behavior is `none`, i.e. layers are stored in tar format with no compression. In testing, this was found to be a _lot_ faster with minimal effect on modelkit size. 

In addition, the `gzip-fastest` option is added as testing showed that it could decrease pack time by a factor of 2-10x with minimal loss in compression ratio.

### Linked issues
Closes #257

### DIFF
--- a/pkg/cmd/list/remote.go
+++ b/pkg/cmd/list/remote.go
@@ -89,7 +89,7 @@ func listImageTag(ctx context.Context, repo registry.Repository, ref *registry.R
 	if err := json.Unmarshal(manifestBytes, manifest); err != nil {
 		return nil, fmt.Errorf("failed to parse manifest: %w", err)
 	}
-	if manifest.Config.MediaType != constants.ModelConfigMediaType {
+	if manifest.Config.MediaType != constants.ModelConfigMediaType.String() {
 		return nil, nil
 	}
 
--- a/pkg/cmd/pack/cmd.go
+++ b/pkg/cmd/pack/cmd.go
@@ -122,11 +122,8 @@ func (opts *packOptions) complete(ctx context.Context, args []string) error {
 		opts.modelRef = repo.DefaultReference()
 	}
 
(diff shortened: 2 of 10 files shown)


gorkem (approved):
LGTM

PR #301, opened 2024-05-08
Pack tarballs using linux path separators on windows

### LINKED ISSUE #291: Model layers created on Windows do not unpack correctly on Linux/MacOS
**Describe the bug**
Modelkits packed on Windows end up using tarballs with windows-style paths (i.e. using `\` as the path separator). When unpacking such models on a non-Windows machine (where `/` is the path separator), directories are not detected and result in filenames instead of directories, e.g. `my-model\my-file.bin` as a filename)

**To Reproduce**
1. Pack model on Windows where some layer refers to a directory that has other directories in it
2. Push model to some registry
3. Unpack model on Linux/MacOS

**Version**
nightly

**Additional context**
N/A

amisevsk:
### Description
Store paths in layer tarballs using linux-style paths (i.e. using separator `/`) on Windows to avoid issues unpacking windows-packed modelkits on Linux/MacOS

### Linked issues
Closes #291

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 .vscode
 kit
+kit.exe
 dist/
 pkg/lib/harness/ui.tar.gz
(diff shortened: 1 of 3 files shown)


gorkem (approved):
lgtm

PR #315, opened 2024-05-14
Update installation guide

bmicklea:
Simplified installation guide

### DIFF
--- a/docs/src/docs/cli/installation.md
+++ b/docs/src/docs/cli/installation.md
@@ -15,7 +15,7 @@ This page includes instructions for:
 
 There are two generations of Mac hardware, if you aren't sure which you have [check here](https://www.sweetwater.com/sweetcare/articles/intel-based-mac-or-mac-with-apple-silicon/#:~:text=Choose%20About%20This%20Mac.,a%20Mac%20with%20Apple%20silicon.&text=As%20of%20this%20writing%2C%20Apple,have%20an%20Intel%2Dbased%20Mac.).
 
-1. <a href="https://github.com/jozu-ai/kitops/releases/download/v0.2.3/kitops-darwin-arm64.zip"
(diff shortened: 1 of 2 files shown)


javisperez (approved):

docs/src/docs/cli/installation.md
```diff
@@ -2,101 +2,144 @@
...
 
-# Installation Guide for `kit` CLI
+# Installing Kit
 
-## Installation from GitHub Releases
-Welcome to the installation guide for the `kit`! This guide is designed to help you quickly and easily install the `kit` on your machine.
+This page includes instructions for:
 
-### Step 1: Downloading the `kit`
+* Installing from [MacOS packages](#🍎-macos-package-download)
```
gorkem, line 9: Should we call these $Platform Installation

docs/src/docs/cli/installation.md
```diff
@@ -2,101 +2,144 @@
...
+* Installing from [Windows packages](#🪟-windows-package-download)
+* Installing from [Linux packages](#🐧-linux-package-download)
+* Building from [source code](#🛠️-install-from-source)
 
-To begin, you will need to download the latest version of `kit`. You can find the most recent release on the official GitHub releases page:
+## 🍎 MacOS Package Download
```
gorkem, line 14: This is not just download it also has instruction to install. I think the heading should reflect that.

gorkem (changes requested):

docs/src/docs/cli/installation.md
bmicklea: Sorry, I don't understand what you mean - do you mean the title should be `MacOS installation` rather than `Installing from MacOS packages`?

gorkem: yes, and same for the following platforms. Windows, Linux

amisevsk (approved):

gorkem (approved):
:shipit:

PR #329, opened 2024-05-28
Chunked repo push

### LINKED ISSUE #328: Chunked uploads for large layers
Kit currently attempts to upload modelkit layers in a single HTTP PUT request. This can lead to unreliability if there is a network issue at some point during the (often long) upload. In addition, uploading large amounts of data in a single request can cause issues on the server side, if requests are buffered. 

The OCI distribution spec defines and API [1] for uploading blobs in multiple chunks rather than as a single monolithic upload. Since layers in modelkits can potentially be very large, it would be more reliable to upload these layers in a chunked format according to the distribution spec. 

[1] - https://github.com/opencontainers/distribution-spec/blob/main/spec.md#push

Implement chunked uploads as defined by the OCI distribution spec. 

N/A

**Additional context**
Additionally, podman and docker both support streaming data to a repo via a single PATCH request using `transfer-encoding: chunked`, though this is not formalized in the OCI spec.

amisevsk:
### Description
Implements chunked blob uploads according to the [OCI distribution spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#push). There are still a few small details to implement, but this should be enough to push large files up to dockerhub, for example.

Currently, kit will upload layers <100MiB monolithically (POST to initiate upload, single PUT to upload file). If the layer is >100MiB, kit will slice the layer into 100MiB chunks and upload them according to the spec.

Kit currently does not support streaming data into the repo via PATCH, as is used in docker and podman.

### Linked issues
Closes #328


pkg/lib/repo/repository.go
```diff
@@ -0,0 +1,270 @@
...
+
+	return blobLocation.String(), nil
+}
+
+// uploadBlobChunked performs a chunked blob upload as per the distribution spec. The blob is divided into chunks of maximum 100MiB
+// in size and uploaded sequentially through PATCH requests. Once entire blob is uploaded, a PUT request marks the upload as complete.
+// Note that the distribution spec 1) requires blobs to uploaded in-order, and 2) does not have a way of specifying maximum blob
+// size.
+func (r *Repository) uploadBlobChunked(ctx context.Context, location *url.URL, postResp *http.Response, expected ocispec.Descriptor, content io.Reader) (string, error) {
+	// TODO: Handle 'OCI-Chunk-Min-Length' header in post response
```
gorkem, line 153: We can not leave this as todo, ghcr push for instance fails for me with `failed to copy to remote: requested_range_not_satisfiable: the request body is too large and exceeds the maximum permissible limit of 4.00MiB`

gorkem (changes requested):

pkg/lib/repo/repository.go
amisevsk: Hm, that's a little strange, since this TODO covers _minimum_ length. The OCI spec has no way to specify _maximum_ length.

Does ghcr really limit chunks to 4.00 MiB?

gorkem: That is right... this is the min.

gorkem: Checked again it gets a 416 from ghcr with the detail above.

amisevsk: Checking the response headers, ghcr.io says nothing to tell us a maximum size :upside_down_face: 

I'm going to special-case it for now.

amisevsk: Added https://github.com/jozu-ai/kitops/pull/329/commits/46f3c7ccb6cbf37f98b5d21050e49f23022e16dd to let us override per-repo, as this seems to be the only way to do this. This effectively reverts to the old way for ghcr.io only.

Checked with podman, and they use `transfer-encoding: chunked` + patch to upload large layers in one go, so we can implement that in the future and potentially use it instead.

pkg/lib/repo/errors.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	body, err := io.ReadAll(io.LimitReader(resp.Body, 8*1024))
+	if err != nil {
+		respErr.ExtraMessage = fmt.Sprintf("Failed to read response body: %s", err)
+		return respErr
+	}
+
+	if err := json.Unmarshal(body, &respErr); err != nil {
+		respErr.ExtraMessage = fmt.Sprintf("response body: %s", string(body))
```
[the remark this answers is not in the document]
amisevsk: The reason I wasn't passing the decode error is that the OCI spec doesn't force a JSON response, only specifies what a JSON response must look like. It's acceptable to return e.g. plain text as the body.

Reworked this section to only try to parse JSON if the content-type is application/json

amisevsk:
Rebased to add license headers to files, which I had forgotten.

pkg/lib/repo/errors.go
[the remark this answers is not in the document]
amisevsk: Maybe also worth noting that `json.Unmarshal` only returns an error if the input isn't valid JSON; it doesn't do any validation, so a JSON response that doesn't match the struct will not report any errors, just an empty struct.

pkg/lib/repo/upload-format.go
```diff
@@ -0,0 +1,43 @@
...
+		// ghcr.io returns 416 is a PATCH has Content-Length greater than 4.0 MiB for some reason
+		// Transfer-Encoding: chunked is supported by the registry, but not implemented yet.
+		return uploadMonolithicPut
+	}
+	// No matches above, use heuristic
+	if size < 100<<20 {
+		return uploadMonolithicPut
+	} else {
+		return uploadChunkedPatch
+	}
```
gorkem, line 42: Should this just be `default:` for the case above?

gorkem (approved):

pkg/lib/repo/upload-format.go
amisevsk: It's equivalent, I did it this way to reduce nesting (switch -> case default -> if/else)

PR #335, opened 2024-05-31
Quick start and next steps updates

### LINKED ISSUE #313: Docs: Using Cosign with KitOps
You can use CoSign or similar to sign ModelKits - documentation should explain (at a high level) why you'd want to do that and how to do it.

### LINKED ISSUE #323: Move `remove` instructions into Quick Start
**Describe the bug**
It's relatively easy to fat finger the tag when creating your ModelKit, but when you're going through the Quick start you don't know how to use remove. Remove should be explained to when you're doing your first pack / list instead of in the Next Steps which people may not read. Also OCI registry isn't a commonly understood term outside DevOps - would be good to demystify it.

**To Reproduce**
* Move examples to GitHub for registry
* Explain that OCI registry == conatiner registry or GitHub registry packages

brett-hodges:
### Description
Added optional `remove` instructions to the Quick Start guide 
Replaced references to docker.io with ghcr.io for consistency in the Quick Start guide
Added cosign to Next Steps

### Linked issues

closes #313 
closes #323

### DIFF
--- a/docs/src/docs/next-steps.md
+++ b/docs/src/docs/next-steps.md
@@ -1,10 +1,18 @@
 # Next Steps with Kit
 
 In this guide you'll learn how to:
+* Sign your ModelKit
 * Make your own Kitfile
 * The power of `unpack`
 * Read the Kitfile or manifest from a ModelKit
 * Tag ModelKits and keep your registry tidy
+
+## Signing your ModelKit
+
+Because ModelKits are OCI artifacts, they can be signed the same way you sign your other containers.
+
+If you need a quick way to sign your containers, check out [Cosign](https://docs.sigstore.dev/signing/quickstart).
+
 
 ## Making your own Kitfile
 
--- a/docs/src/docs/next-steps.md
+++ b/docs/src/docs/next-steps.md
@@ -9,7 +9,7 @@ In this guide you'll learn how to:
 
 ## Signing your ModelKit
 
-Because ModelKits are OCI artifacts, they can be signed like any other OCI artifact (you may already sign your containers, for example.)
+Because ModelKits are OCI artifacts, they can be signed like any other OCI artifact (you may already sign your containers, for example).
 
 If you need a quick way to sign a ModelKit you can follow the same instructions as for a container, using something like [Cosign](https://docs.sigstore.dev/signing/quickstart).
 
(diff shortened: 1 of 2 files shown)


docs/src/docs/next-steps.md
```diff
@@ -1,11 +1,19 @@
...
 In this guide you'll learn how to:
+* Sign your ModelKit
 * Make your own Kitfile
 * The power of `unpack`
 * Read the Kitfile or manifest from a ModelKit
 * Tag ModelKits and keep your registry tidy
 
+## Signing your ModelKit
+
+Because ModelKits are OCI artifacts, they can be signed the same way you sign your other containers.
```
bmicklea, line 12: Let's be careful about making it sound like ModelKits are containers. Saying something like "...they can be signed like any other OCI artifact (you may already sign your containers for example)" would be better.

docs/src/docs/next-steps.md
```diff
@@ -1,11 +1,19 @@
...
+
+If you need a quick way to sign your containers, check out [Cosign](https://docs.sigstore.dev/signing/quickstart).
```
bmicklea, line 14: Same here - say something more like "...way to sign a ModelKit you can follow the same instructions as for a container using something like [Cosign]"

bmicklea (changes requested):
I'm hyper-sensitive about if and how we use "containers" in connection with KitOps. People have thought that we were just a replacement for the Docker CLI which kills our value so my comments are mostly about how to reword things to either get away from saying container or make the distinction between conatiners and ModelKits clearer...

docs/src/docs/next-steps.md
brett-hodges: I made the change you suggested here.

docs/src/docs/next-steps.md
brett-hodges: I made the change you suggested here, too.

bmicklea (approved):

docs/src/docs/next-steps.md
```diff
@@ -1,11 +1,19 @@
...
 In this guide you'll learn how to:
+* Sign your ModelKit
 * Make your own Kitfile
 * The power of `unpack`
 * Read the Kitfile or manifest from a ModelKit
 * Tag ModelKits and keep your registry tidy
 
+## Signing your ModelKit
+
+Because ModelKits are OCI artifacts, they can be signed like any other OCI artifact (you may already sign your containers, for example.)
```
javisperez, line 12: Not a blocker, but having the period inside the parenthesis is making my eye twitch haha.

javisperez (approved):

PR #339, opened 2024-06-06
Add different log-levels to output; name per-request trace logging for registry operations

### LINKED ISSUE #338: Improve logging output with more information
Deeper debug/trace logging in kit is currently lacking. It would be ideal to support additional logging (and log levels) to allow e.g. logging requests to remote registries. An example of what this could look like is `podman`'s `--log-level` flag, which at a debug/trace level will output a lot of useful information in debugging issues.

**Additional context**
N/A

amisevsk:
### Description
Add support for multiple log levels (`error`, `warn`, `info`, `debug`, and `trace`) with a corresponding flag `--log-level`. The old `-v | --verbose` flag is left in place for convenience. Currently, the only 'trace' level logging is per-request info while pushing/pulling a registry. Also, if the log-level is not `info`, we output a prefix indicating log level for each line, to make it a little more readable.

If the log-level is 'info", the output is unchanged.

### Linked issues
Closes #338

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -36,6 +36,7 @@ Find more information at: http://kitops.ml`
 type rootOptions struct {
 	configHome   string
 	verbose      bool
+	loglevel     string
 	progressBars string
 }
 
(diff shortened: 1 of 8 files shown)


amisevsk:
There's some oddness in the color selections -- I'd ideally like `[TRACE]` to be a dim grey, but some color themes (looking at you, solarized on gnome terminal) use the same color for background and "dark grey" making it invisible.

gorkem (approved):
I kind of wish we could implement a `Logger` struct that encapsulate and abstract away this.

amisevsk:
> I kind of wish we could implement a Logger struct that encapsulate and abstract away this.

I agree, though at this point, that's basically what we've done -- the struct would just encapsulate what are now package globals :shrug:

I can open a follow-up PR to clean it up a little. The cleanest in-package solution would be to do away with `info`, `debug`, `error` functions and just use `log(LEVEL, ...)` but that makes every invocation a little messier

PR #382, opened 2024-06-25
Fix issue when using relative paths for KITOPS_HOME

amisevsk:
### Description
Since kit sometimes changes directory while packing/unpacking, using a relative path for KITOPS_HOME (or the --config parameter) would lead to the CLI using an unexpected location -- for example, running

```
  KITOPS_HOME=./my-kitops-storage
  kit pack -t my-model ./my-model-path
```

would lead to kit storing the model in `./my-model-path/my-kitops-storage` instead of the expected `./my-kitops-storage`.

Instead, we take the absolute path for the parameter at the beginning and use that.

### Linked issues
Closes https://github.com/jozu-ai/kitops/issues/381

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -5,7 +5,9 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 	"os"
+	"path/filepath"
 
 	"kitops/pkg/cmd/dev"
 	"kitops/pkg/cmd/info"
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -117,13 +119,21 @@ func Execute() {
 func getConfigHome(opts *rootOptions) (string, error) {
 	if opts.configHome != "" {
 		output.Debugf("Using config directory from flag: %s", opts.configHome)
-		return opts.configHome, nil
+		absHome, err := filepath.Abs(opts.configHome)
```
gorkem, line 122: this is repeated. Should it be a nested helper function something like.

```
getAbsPath := func(path, description string) (string, error) {
		absPath, err := filepath.Abs(path)
		if err != nil {
			return "", fmt.Errorf("failed to get absolute path for %s: %w", description, err)
		}
		return absPath, nil
	}
```

gorkem (approved):

cmd/root.go
amisevsk: I feel like this level of repetition is just annoyingly inherent to golang (and trying to reduce it would just complicate the code). 

The only part we're really repeating is handling the error, and I think I prefer to more clearly distinguish the message between "we couldn't figure out the path you passed via `--context`" vs the more subtle "`KITOPS_HOME` is set somewhere and that's messing things up."

e.g.
```go
getAbsPath := func(...)

absHome, err := getAbsPath(opts.configHome, "--config parameter")
if err != nil {
    return err
}
```

vs.

```go
absHome, err := filepath.Abs(path)
if err != nil {
    return fmt.Errorf("<message>")
}
```

PR #385, opened 2024-06-26
Add support for arbitrary parameters in models

### LINKED ISSUE #366: Add Ability to attach metadata to artifacts
It should be possible to attach metadata to artifacts (models, datasets) that may be coming from third-party tools and store them in the kitfile. This metadata should be easily separable and organized for specific purposes or tools.

I would like to add key-value pairs as parameters to models and datasets that are easily separated for specific purposes, tools, etc. This will allow us to include metadata from different sources in a clear and organized manner.

```YAML
model:
    name: ModelName
    path: models/model.h5
    parameters:
      test-parameters:
         learning_rate: 0.001
        epochs: 100
        batch_size: 32
      mlflow:
        created_at: 2023-02-01
        updated_at: 2023-02-10
        mlflow_specific_key: mlflow_value
      custom-tool:
        version: 1.2.1
        tested: true
```

Similarly for datasets

```YAML
datasets:
  - name: DatasetName
    path: data/dataset.csv
    metadata:
      source_info:
        source: data_source_name
        size: 100MB
        format: CSV
        features: 10
        records: 1000
      tool1:
        created_at: 2023-01-01
        updated_at: 2023-01-15
      tool2:
        checksum: abcdef123456
        validated: true
```

An alternative could be to use a more generic key-value structure without separating by specific tools or purposes, but this would reduce clarity and make it harder to manage and interpret the metadata.

**Additional context**

amisevsk:
### Description
Add ability to store arbitrary additional data in a Kitfile's model, e.g.

```yaml
model:
  - name: my-model
    path: ./model.bin
    parameters:
      <any valid yaml>
```

This comes with some caveats in serialization:
* Numerical fields will be converted to decimal representation, so `0xFF` will be come `255`, `1.2e+3` will become `1200`
* Strings will be serialized ignoring flow/etc.
* Anything yaml-specific (i.e. not included in JSON) is unsupported
* Maps will have their keys sorted alphabetically (`{b: b, a: a} -> {a: a, b: b}`)

The net effect of this is that the unpacked model may have a (superficially) different Kitfile than the original, but both Kitfiles will pack to the same digest.

### Linked issues
Closes #366

### DIFF
--- a/pkg/artifact/kit-file.go
+++ b/pkg/artifact/kit-file.go
@@ -66,6 +66,13 @@ type (
 		Training    *Training   `json:"training,omitempty" yaml:"training,omitempty"`
 		Validation  *Validation `json:"validation,omitempty" yaml:"validation,omitempty"`
 		Parts       []ModelPart `json:"parts,omitempty" yaml:"parts,omitempty"`
+		// Parameters is an arbitrary section of yaml that can be used to store any additional
+		// data that may be relevant to the current model, with a few caveats:
+		//  * Only a json-compatible subset of yaml is supported
+		//  * Strings will be serialized without flow parameters, etc.
+		//  * Numbers will be converted to decimal representations (0xFF -> 255, 1.2e+3 -> 1200)
+		//  * Maps will be sorted alphabetically by key
+		Parameters any `json:"parameters,omitempty" yaml:"parameters,omitempty"`
 	}
 
 	ModelPart struct {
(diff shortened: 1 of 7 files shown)


gorkem (commented):
Can you fix the build failure

amisevsk:
Ah, forgot a license header

gorkem (approved):

PR #389, opened 2024-07-02
Slim kitfile spec

amisevsk:
### Description
Trim down the kitfile spec for simplicity (removed fields can be re-added later if we find a need for them):
* ~~Remove license field from `model`, `dataset`, and `code`, leaving only the top-level package license~~
* Remove dataset references from model; since we only have one model now, datasets are assumed to apply to it
* Remove processing field from datasets
* Minor renames of internal fields to match their names in the Kitfile (e.g. `TrainedModel -> Model`)

This PR also changes how Kitfiles are parsed to print an error if an unknown field is present:
```
❯ kit pack -t test .
[ERROR] Failed to pack model kit: yaml: unmarshal errors:
  line 2: field manfiestVersion not found in type artifact.KitFile
```
(This could be cleaned up but the yaml package does not give us many easy options to do so right now)

### Linked issues
Closes https://github.com/jozu-ai/kitops/issues/239


gorkem (approved):
We need docs update to accompy this but otherwise looks good.

amisevsk:
Turns out I had some copy-paste typos in the integration tests and this caught them, so we're already ahead :D

gorkem:
@amisevsk We have been discussing this with @bmicklea and he made the point that it is actually the top level license that is creating ambiguity and we should keep the licenses for code, model and datasets and deprecate the package level license.

amisevsk:
I was kind of heading the other direction, I was considering turning the package-level license field into a list, so you could have
```yaml
package:
  license: ["MIT", "CC BY-SA 4.0"]
```
or similar. At the end of the day, unpacking any of the layers _should_ unpack a LICENSE file as well. 

I'll update it to revert the license change (and then re-update https://github.com/jozu-ai/kitops/pull/395 to match)

amisevsk:
Alright all license changes are reverted. We want licenses on models/datasets/code sections, and have to keep the package-level license field too since most existing modelkits use it.

amisevsk:
Docs changes are in https://github.com/jozu-ai/kitops/pull/395

PR #392, opened 2024-07-04
Allow specifying -vv to set log-level to trace

amisevsk:
### Description
Allow -v flag to be specified multiple times, so that

  -v  : verbosity 1 (equivalent to --log-level=debug)
  -vv : verbosity 2 (equivalent to --log-level=trace)

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -5,6 +5,7 @@ package cmd
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


gorkem (commented):
works like a charm

gorkem (approved):

PR #407, opened 2024-07-12
Don't fail remote list operations if repo contains non-modelkits

### LINKED ISSUE #403: Kit list fails if remote repo has artifacts that are not ModelKits
**Describe the bug**
When invoking a `kit list` command on a repo if the repo has artifacts (like signatures) that are not ModelKits the kit list command fails. I suspect the culprit  is this line [line](https://github.com/amisevsk/kitops/blob/bec85c4b8f3d64731451020aaedc184dbedebb4b/pkg/lib/repo/util/reference.go#L202)

**To Reproduce**
1. ```kit list jozu.ml/jozu/gemma -vvv```

**Version**

**Additional context**

amisevsk:
### Description
Normally, attempting to read a manifest from a repo fails if that descriptor does not refer to a modelkit. However, for the `kit list` operation, we want to just ignore such errors to allow listing all modelkits in repositories that may also contain e.g. docker images or other OCI artifacts.

### Linked issues
Closes #403

### DIFF
--- a/pkg/cmd/list/list.go
+++ b/pkg/cmd/list/list.go
@@ -5,6 +5,7 @@ package list
 
 import (
 	"context"
+	"errors"
 	"sort"
 
 	"kitops/pkg/lib/constants"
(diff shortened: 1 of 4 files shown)


gorkem (approved):
lgtm

gorkem:
The new `errors.go` is missing the license header

PR #410, opened 2024-07-15
jozu hub compatibility

bmicklea:
**This should not be merged until after the Jozu Hub is launched**

Updated docs to simplify pulling ModelKits by grabbing them from Jozu Hub.

Since we can only pull from Jozu Hub, not push, I left things like `kit login` and `kit push` pointing to `ghcr.io`. Once Jozu Hub can accept pushes we'll swap those over so new and existing users can see the ModelKits with more metadata and detail in the Hub.


javisperez:
@bmicklea maybe this could be turned into a Draft PR, to avoid accidentally merging it?

bmicklea:
We can merge this now that the jozu.ml is working properly.

javisperez (approved):

PR #418, opened 2024-07-18
Show update notifications when a new version of kit is available

### LINKED ISSUE #365: Add notification for CLI users for new version
Right now there's not push notification style way for a user to know that a new CLI version is available and that they're behind.

Some kind of warning or notification when a new CLI version is cut.

amisevsk:
### Description
Show an update notification in the CLI if a newer version is available on GitHub:

```
Note: A new version of Kit is available! You are using Kit v0.2.2. The latest version is v0.3.0.
      To see a list of changes, visit https://github.com/jozu-ai/kitops/releases/tag/v0.3.0
      To disable this notification, use 'kit version --show-update-notifications=false'
```

Notifications can be disabled by running `kit version --show-update-notifications=false` (use `=true` to re-enable). 

Errors in checking for an update are delegated to debug-level logs and not shown to the user by default. 

To test this PR, you will need to build kit with at least `-ldflags="-s -w -X kitops/pkg/lib/constants.Version=${version}"`

Note: I had to put the enable/disable flag under the version subcommand, as trying to include it in the root command would either not run it or would break the default help text.

### Linked issues
Closes #365

### DIFF
--- a/pkg/cmd/version/version.go
+++ b/pkg/cmd/version/version.go
@@ -18,6 +18,7 @@ package version
 
 import (
 	"kitops/pkg/lib/constants"
+	"kitops/pkg/lib/update"
 	"kitops/pkg/output"
 
 	"github.com/spf13/cobra"
@@ -32,16 +33,35 @@ the version was built from, the build time, and the version of Go it was
 compiled with.`
 )
 
+const versionNotifFlag = "show-update-notifications"
+
+type versionOpts struct {
+	shouldShowNotifications bool
+}
+
 func VersionCommand() *cobra.Command {
+	opts := &versionOpts{}
 
 	cmd := &cobra.Command{
 		Use:   "version",
 		Short: shortDesc,
 		Long:  longDesc,
 		Run: func(cmd *cobra.Command, args []string) {
-			output.Infof("Version: %s\nCommit: %s\nBuilt: %s\nGo version: %s\n", constants.Version, constants.GitCommit, constants.BuildTime, constants.GoVersion)
+			if cmd.Flags().Changed(versionNotifFlag) {
+				configHome, ok := cmd.Context().Value(constants.ConfigKey{}).(string)
+				if !ok {
+					output.Fatalln("default config path not set on command context")
+				}
+				if err := update.SetShowNotifications(configHome, opts.shouldShowNotifications); err != nil {
+					output.Fatalln(err)
+				}
+			} else {
+				output.Infof("Version: %s\nCommit: %s\nBuilt: %s\nGo version: %s\n", constants.Version, constants.GitCommit, constants.BuildTime, constants.GoVersion)
+			}
 		},
 	}
+	cmd.Flags().BoolVar(&opts.shouldShowNotifications, versionNotifFlag, false, "Enable or disable update notifications for the Kit CLI")
+
 	return cmd
 }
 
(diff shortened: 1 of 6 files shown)


pkg/lib/constants/consts.go
```diff
@@ -34,14 +34,15 @@ const (
...
-	HarnessLogFile      = "harness.log"
+	// Modelkits are stored in $KITOPS_HOME/storage/ and
+	// credentials are stored in $KITOPS_HOME/credentials.json
+	DefaultConfigSubdir               = "kitops"
+	StorageSubpath                    = "storage"
+	CredentialsSubpath                = "credentials.json"
+	HarnessSubpath                    = "harness"
+	HarnessProcessFile                = "process.pid"
+	HarnessLogFile                    = "harness.log"
+	UpdateNotificationsConfigFilename = "disable-update-notifications"
```
gorkem, line 45: Should we just have a `kitconfig.json` where this is just a key?

pkg/cmd/version/version.go
```diff
@@ -32,16 +33,35 @@ the version was built from, the build time, and the version of Go it was
...
 		Run: func(cmd *cobra.Command, args []string) {
-			output.Infof("Version: %s\nCommit: %s\nBuilt: %s\nGo version: %s\n", constants.Version, constants.GitCommit, constants.BuildTime, constants.GoVersion)
+			if cmd.Flags().Changed(versionNotifFlag) {
+				configHome, ok := cmd.Context().Value(constants.ConfigKey{}).(string)
+				if !ok {
+					output.Fatalln("default config path not set on command context")
+				}
+				if err := update.SetShowNotifications(configHome, opts.shouldShowNotifications); err != nil {
+					output.Fatalln(err)
+				}
```
gorkem, line 57: You can possibly move the code for handling the flag to its own function.

pkg/lib/update/update.go
```diff
@@ -0,0 +1,133 @@
...
+		return
+	}
+	defer resp.Body.Close()
+
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		output.Debugf("Failed to read GitHub response body: %s", err)
+		return
+	}
+	info := &ghReleaseInfo{}
```
gorkem, line 83: The code for retrieving the release info could be refactored into a function for clarity.

gorkem (commented):
Is it possible to print the new release note after the output of the command? Is that a better experience?

pkg/lib/constants/consts.go
amisevsk: That's a much bigger task :)

If we have a kitconfig.json, I'd expect to be able to almost configure everything there (e.g. something like viper). What that looks like is pretty nebulous.

pkg/cmd/version/version.go
amisevsk: I could, but it would only reduce this section from 7 lines to 3 (call complete and check the error). This way it's simple to suppress the default output for `kit version` when the option is set.

amisevsk:
> Is it possible to print the new release note after the output of the command? Is that a better experience?

We could potentially put it in `postRun`, but in my experience most applications print update notifications at the top before showing you the output you want. We don't want to be intrusive.

pkg/lib/constants/consts.go
gorkem: let's file an issue for it

amisevsk: Filed https://github.com/jozu-ai/kitops/issues/419

pkg/lib/update/update.go
amisevsk: Done

gorkem (approved):

PR #420, opened 2024-07-19
Fix garbage collection for pulled modelkits

amisevsk:
### Description
When we pull a modelkit from a remote, we need to ensure its manifest descriptor is added to both the local index and the shared (non-scoped) index. Otherwise, garbage collection will fail -- `kit remove <modelkit>` will remove the manifest from the index and it won't appear in `kit list`, but the blobs referenced by the manifest will not be cleaned up there is no manifest to delete in the shared OCI index.

Calling delete on a non-existent manifest in the store will not trigger an error, so this issue was a little hard to notice.

### Linked issues
N/A


gorkem (approved):
LGTM

PR #428, opened 2024-07-30
Updated prebuilt-modelkits.md to remove references to invalid datasets

brett-hodges:
Removed invalid ModelKits that were meant to be included only for internal testing.


bmicklea (approved):
Thanks for noticing this Brett - approved

PR #435, opened 2024-08-02
Update ModelKit diagram

bmicklea:
Also add in a call-out for EU AI Act


docs/src/docs/modelkit/ModelKit_chart.svg
gorkem, line 1: In this diagram it looks like the hyperparameters and key value-pairs, are part of docs. However, these are model parameters.  For the docs we can list Model Card, Dataset Metadata probably

gorkem (changes requested):
Diagram probably needs another iteration

docs/src/docs/modelkit/ModelKit_chart.svg
gorkem: Why store the key-value pairs and JSON as doc and not code? On the kitfile, we do have `model.parameters` which can be used to store key-value pairs.

bmicklea: Just saw your docs update in the other PR. I misunderstood how this worked. I'll rework the diagram.

gorkem: Why are key-value pairs and JSON doc and not code? On the kitfile, we do have `model.parameters` which can be used to store key-value pairs.

bmicklea: Why do you want it to be so specific? Could it not be metadata about the code/model interactions or about testing methodology? Or about compliance factors?

gorkem: It could but when creating SBOMs. two source of metadata will be important Model Cards and Dataset Metadata. I thought it would be good to refer to those.

gorkem (approved):

PR #436, opened 2024-08-02
Update model parameters to kitfile spec

gorkem:
Updates the kitfile spec, which should also update the generated docs.


bmicklea (approved):
LGTM

PR #446, opened 2024-08-15
Enable unpack to filter for specific layers instead of all layers of a type

amisevsk:
### Description
This PR adds a flag `--filter` (`-f` for short) that can be used to filter which parts of a Kitfile are unpacked in a `kit unpack` operation. In general, the flag takes the format 
```
<types>:<filters>
```
where `<types>` is a comma-separate list of sections in the kitfile to unpack (e.g. `model,datasets,kitfile`), and `<filters>` is an optional comma-separated list of filters to apply to those types. To unpack only models and datasets, you would use the command
```
kit unpack <model> --filter=model,datasets
```
To unpack only the `training-data` dataset, you would use
```
kit unpack <model> --filter=datasets:training-data
```

Matching for `<filters>` is done based on name (e.g. dataset/modelpart name) or on path (since not all fields have an ID and names are optional). Currently, matching on paths is exact and not anything fancy.

The old flags (`--model`, `--datasets`, etc.) are still present for compatibility and marked as deprecated (for removal in a few releases). They currently are converted to basic filters to work within the new system.

### Linked issues
Closes https://github.com/jozu-ai/kitops/issues/391

### DIFF
--- a/pkg/cmd/unpack/cmd.go
+++ b/pkg/cmd/unpack/cmd.go
@@ -41,13 +41,24 @@ filesystem. By default, it attempts to find the modelkit in local storage; if
 not found, it searches the remote registry and retrieves it. This process
 ensures that the necessary components are always available for unpacking,
 optimizing for efficiency by fetching only specified components from the
-remote registry when necessary`
+remote registry when necessary.
+
+The content that is unpacked can be limited via the --filter (-f) flag. For example,
+use
+    --filter=model
+to unpack only the model, or 
+    --filter=datasets:my-dataset
+to unpack only the dataset named 'my-dataset'. Filtering for specific code/datasets/docs
+will match on either name or the path of the field.`
 
 	example = `# Unpack all components of a modelkit to the current directory
 kit unpack myrepo/my-model:latest -d /path/to/unpacked
 
 # Unpack only the model and datasets of a modelkit to a specified directory
-kit unpack myrepo/my-model:latest --model --datasets -d /path/to/unpacked
+kit unpack myrepo/my-model:latest --filter=model,datasets -d /path/to/unpacked
+
+# Unpack only the dataset named "my-dataset" to the current directory
+kit unpack myrepo/my-model:latest --filter=datasets:my-dataset
 
 # Unpack a modelkit from a remote registry with overwrite enabled
 kit unpack registry.example.com/myrepo/my-model:latest -o -d /path/to/unpacked`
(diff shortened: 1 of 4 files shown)


pkg/cmd/unpack/cmd.go
```diff
@@ -41,27 +41,42 @@ filesystem. By default, it attempts to find the modelkit in local storage; if
...
+    --filter=model
+to unpack only the model, or 
+    --filter=datasets:my-dataset
+to unpack only the dataset named 'my-dataset'. Filtering for specific code/datasets/docs
+will match on either name or the path of the field.`
 
 	example = `# Unpack all components of a modelkit to the current directory
 kit unpack myrepo/my-model:latest -d /path/to/unpacked
 
 # Unpack only the model and datasets of a modelkit to a specified directory
```
gorkem, line 57: Can we have an example that shows multiple filters. Also a better explanation for the cases  where multiple filter contents can be used.

gorkem (changes requested):

pkg/cmd/unpack/filter.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	return []filterConf{filter}
+}
+
+func filterToMediaBaseType(filterType string) (string, error) {
+	switch filterType {
+	case "kitfile":
+		return constants.ConfigType, nil
+	case "datasets":
+		// annoyingly, the mediatype is dataset, but for the filter we want the plural
```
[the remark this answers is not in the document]
amisevsk: I think I prefer `datasets` -- while it doesn't match the mediatype, it does match the corresponding field in the Kitfile, which is going to be more familiar to users.

As an aside, I considered going for `dataset` here to make matching to a mediatype easier, but it reads kind of weird to me. The other filters, `code` and `docs`, work because they're weird words (mass/collective nouns? I'm bad at grammar but we have to consistently call them "code layers" or something else)

pkg/cmd/unpack/cmd.go
amisevsk: Hm, it is getting long and complicated. Maybe it makes sense to have a separate "Filtering" section in the help page. I'll come up with something.

amisevsk: I couldn't add a section (the whole desc is indented so it looks weird) but I've expanded this description and added more examples.

pkg/cmd/unpack/filter.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func (fc *filterConf) matches(baseType, field string) bool {
+	var matchesType, matchesField bool
+	for _, t := range fc.baseTypes {
+		if t == baseType {
+			matchesType = true
+			break
+		}
+	}
```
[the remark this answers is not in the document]
amisevsk: Good point, reworked this bit.

gorkem (commented):
There are failing tests

amisevsk:
Aha, the automated tests coming in handy once again. Unpacking parents was using the now-ignored unpackConf.

gorkem (approved):

PR #452, opened 2024-08-19
Fix indefinite hang when unpacking over existing files

### LINKED ISSUE #448: Unpack hangs when files already exists and no overwrite flag is given
**Describe the bug**
When unpacking for flag --code, if the code file already exists and does not provide -o . The unpacking hangs with progress bar, but does not prompt error. 

**Version**
Version: 0.3.2
Commit: 5e99ff38fcd5f4c75880a6516cc9e4f940435c26
Built: 2024-08-02T12:33:31Z
Go version: go1.21.6

**Additional context**
Intel 64bit
NAME="Ubuntu"
VERSION="18.04.5 LTS (Bionic Beaver)"

amisevsk:
### Description
We have to wait for progress bars to finish before returning in unpack. However, if the unpack process encounters an error, the progress bar will never finish, meaning a deferred Wait() call will hang indefinitely.

Instead, we only wait in cases of success, and return the error if one is encountered.

### Linked issues
Closes #448

### DIFF
--- a/pkg/cmd/unpack/unpack.go
+++ b/pkg/cmd/unpack/unpack.go
@@ -200,7 +200,6 @@ func unpackLayer(ctx context.Context, store content.Storage, desc ocispec.Descri
 	var logger *output.ProgressLogger
 	rc, logger = output.WrapReadCloser(desc.Size, rc)
 	defer rc.Close()
-	defer logger.Wait()
 
 	var cr io.ReadCloser
 	var cErr error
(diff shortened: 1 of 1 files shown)


pkg/cmd/unpack/unpack.go
```diff
@@ -221,7 +220,11 @@ func unpackLayer(ctx context.Context, store content.Storage, desc ocispec.Descri
 		return fmt.Errorf("failed to create directory %s: %w", unpackDir, err)
```
gorkem, line 220: Not related to these changes but on line 213 `cErr` is checked for nil but on the `err` is used for return value.

gorkem (approved):

pkg/cmd/unpack/unpack.go
amisevsk: Good catch! Fixed.

PR #453, opened 2024-08-20
Enable building containerized Kit CLI and init container

### LINKED ISSUE #369: init container for ModelKits
Using init containers to prime the inference engine with the model artifacts is a commonly used pattern for running models on a Kubernetes cluster. However, there is currently no trusted, maintained, and secure init container image available that can be easily used by ModelKit users.

Create an init container with provenance and signing capabilities. The container should be configurable to unpack specific parts of the ModelKit to designated locations. The container should do signature verification when signatures are available. We should provide a container for each version of the Kit CLI, starting with the current one. Additionally, comprehensive documentation should be available, including examples of how to use the init container effectively.

* Manually creating and maintaining custom init containers for each use case.
* Using existing third-party init containers that may not meet our security and provenance requirements.

**Additional context**
- For security, we should enable measures such as, a trusted base image, regular updates, vulnerability scanning.
- There should be a container with the same tag as the kit CLI version that would make it convenient to select a container with latest pointing to latest kit CLI version.

amisevsk:
### Description
Add a Dockerfile that just builds the Kit CLI and includes it in a container. While this container can be used to quickly run Kit commands, e.g.

```shell
docker run -it --rm kitimage:latest unpack mymodel -d /home/user/test
```

it is also used as a base image for other images:
* The KServe Dockerfile is largely unchanged and kept in case someone wants to use it
* An additional Kubernetes-style init container is added and makes it easier to run Kit in an init container to unpack a modelkit to some path in a volume. This container also supports verifying signatures using cosign.

### Still TODO:
* Decide which repo to store these images in
* Add CI configurations to build versioned images on release and `next` images for each commit
* Figure out better base images for containers than just alpine.

### Linked issues
Closes #369

### DIFF
--- a/.github/workflows/next-container-build.yaml
+++ b/.github/workflows/next-container-build.yaml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+    paths-ignore: docs/
 
 env:
   REGISTRY: ghcr.io
--- a/.github/workflows/next-container-build.yaml
+++ b/.github/workflows/next-container-build.yaml
@@ -1,0 +1,61 @@
+name: Next CLI container build
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+  INIT_IMAGE_NAME: ${{ github.repository }}-init
+  NEXT_TAG: next
+
+permissions:
+  packages: write
+
+jobs:
+  docker-image-build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@49b3bc8e6bdd4a60e6116a5414239cba5943d3cf    # v3.2.0
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@988b5a0280414f521da01fcc63a27aeeb4b104db  # v3.6.1
+
+      - name: Login to ghcr.io
+        uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567         # v3.3.0
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Checkout
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+      - name: Build and push base Kit container
+        id: build-kit-container
+        uses: docker/build-push-action@5cd11c3a4ced054e52742c5fd54dca954e0edd85    # v6.7.0
+        with:
+          platforms: linux/amd64,linux/arm64
+          push: true
+          context: .
+          file: build/dockerfiles/Dockerfile
+          build-args: |
+            KIT_VERSION=${{ github.ref_name }}
+          tags: |
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.NEXT_TAG }}
+
+      - name: Build and push Kit init container
+        uses: docker/build-push-action@5cd11c3a4ced054e52742c5fd54dca954e0edd85    # v6.7.0
+        with:
+          platforms: linux/amd64,linux/arm64
+          push: true
+          context: build/dockerfiles/init
+          file: build/dockerfiles/init/Dockerfile
+          build-args: |
+            KIT_BASE_IMAGE=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}@${{ steps.build-kit-container.outputs.digest }}
+          tags: |
+            ${{ env.REGISTRY }}/${{ env.INIT_IMAGE_NAME }}:${{ env.NEXT_TAG }}
--- a/.github/workflows/next-container-build.yaml
+++ b/.github/workflows/next-container-build.yaml
@@ -13,6 +13,9 @@ env:
   NEXT_TAG: next
 
 permissions:
+  id-token: write
+  contents: read
+  attestations: write
   packages: write
 
 jobs:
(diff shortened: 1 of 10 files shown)


build/dockerfiles/Dockerfile
```diff
@@ -14,7 +13,7 @@ COPY . .
 RUN \
     CGO_ENABLED=0 go build \
     -o kit \
-    -ldflags="-s -w -X kitops/pkg/lib/constants.Version=${version} -X kitops/pkg/lib/constants.GitCommit=$gitCommit -X kitops/pkg/lib/constants.BuildTime=$(date -u +'%Y-%m-%dT%H:%M:%SZ')"
+    -ldflags="-s -w -X kitops/pkg/lib/constants.Version=${version} -X kitops/pkg/lib/constants.GitCommit=$(git rev-parse --short HEAD) -X kitops/pkg/lib/constants.BuildTime=$(date -u +'%Y-%m-%dT%H:%M:%SZ')"
```
gorkem, line 16: Should we do a build here or download from releases? Downloading would be make it easier for us to correlate.

amisevsk: We could download, but then we'd need an annoying arch switch and it'd be harder to do `next`-style image builds.

I suppose it would be useful to have the build timestamp match up though. Maybe we do both?

build/dockerfiles/release.Dockerfile
gorkem, line 1: I tried to build this locally but it failed with the following.

```
ERROR: failed to solve: process "/bin/sh -c   set -e\n\n  [ -n \"$KIT_VERSION\" ] || { echo \"Build arg 'KIT_VERSION' is required\"; exit 1; }\n\n  export ARCH=\"$(uname -m | sed 's|aarch64|arm64|g')\"\n\n  echo \"Downloading CLI and checksums\"\n  curl -fsSL \\\n    -o kit.tar.gz \"https://github.com/jozu-ai/kitops/releases/download/$KIT_VERSION/kitops-linux-$ARCH.tar.gz\" \\\n    -o checksums.txt \"https://github.com/jozu-ai/kitops/releases/download/$KIT_VERSION/kitops_${KIT_VERSION}_checksums.txt\" \\\n\n  echo \"Checking SHA256 checksum\"\n  CHECKSUM=$(cat checksums.txt | grep \"kitops-linux-$ARCH.tar.gz\" | cut -f 1 -d ' ')\n  echo \"$CHECKSUM kit.tar.gz\" | sha256sum -c\n\n  echo \"Extracting Kit CLI to $(pwd)/extracted/\"\n  mkdir -p ./extracted\n  tar -xzvf kit.tar.gz -C ./extracted/\n" did not complete successfully: exit code: 22
```

build/dockerfiles/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+The Dockerfile in this directory can be used for building a containerized version of the Kit CLI. To build this image manually, run the following command from the root of this repository:
+```bash
+docker build -t kit-cli:latest -f build/dockerfiles/Dockerfile .
+```
+
+The default entrypoint for this image is invoking the `kit` CLI, allowing for quickly running Kit commands without needing to install the CLI:
+```
+docker run --rm ghcr.io/jozu-ai/kit:latest version
+```
```
gorkem, line 11: Should probably have information about release.Dockerfile?

build/dockerfiles/init/entrypoint.sh
```diff
@@ -0,0 +1,35 @@
...
+fi
+
+# Variables for verifying signature via cosign. Can specify either a key to use for
+# verifying or an identity and oidc issuer for keyless verification
+if [ -n "$COSIGN_KEY" ]; then
+  echo "Verifying signature for modelkit $MODELKIT_REF via key"
+  cosign verify --key "$COSIGN_KEY" "$MODELKIT_REF"
+elif [ -n "$COSIGN_CERT_IDENTITY" ] && [ -n "$COSIGN_CERT_OIDC_ISSUER" ]; then
+  echo "Verifying signature for modelkit $MODELKIT_REF"
+  cosign verify \
```
gorkem, line 23: Does this need like an `if` to check if verify was successful and exit non-zero otherwise?

build/dockerfiles/init/entrypoint.sh
```diff
@@ -0,0 +1,35 @@
...
+    "$MODELKIT_REF"
+else
+  echo "Signature verification is disabled"
+fi
+
+echo "Binary version info:"
+kit version
+
+echo "Unpacking modelkit $MODELKIT_REF to $UNPACK_PATH with filter $UNPACK_FILTER"
+kit unpack "$MODELKIT_REF" --dir "$UNPACK_PATH" --filter="$UNPACK_FILTER"
```
gorkem, line 35: Similiarly does this need a success check?

.github/workflows/next-container-build.yaml
```diff
@@ -0,0 +1,87 @@
+name: Next CLI container build
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
```
gorkem, line 8: we can ignore the `docs/` for this flow 
```suggestion
    paths-ignore: docs/
```

gorkem (changes requested):

build/dockerfiles/init/entrypoint.sh
amisevsk: No, because we `set -e` at the top; any failed command will exit with the exit code of that command.

build/dockerfiles/release.Dockerfile
amisevsk: Hm, works for me. Did you include `--build-arg KIT_VERSION=v0.3.3` or similar?

gorkem: I did... but did not have the` v` :) Works now

build/dockerfiles/README.md
amisevsk: Done

gorkem (approved):

PR #454, opened 2024-08-21
Updated Blog Feed + Readme

Jwilliamsr:
Added recent kitops blog posts & a few typos in readme fixed.

PR #455, opened 2024-08-21
Fixed Kit > KitOps references on home page of kitops

Jwilliamsr:


gorkem (changes requested):

javisperez (changes requested):
@Jwilliamsr i think you accidentally re used an stale branch because most of this changes we already approved and merged yesterday. I think you need to close this pr and maybe pull the latest, create a new branch and then PR again because this diff is false.

Check #454

PR #485, opened 2024-09-24
Add to use cases

bmicklea:
* Added a new Level 2 use case for security referencing ModelScan + KitOps
* Made problem more clear (moved up)
* Renamed Level 2 to Level 3


brett-hodges (approved):
Very nice!

PR #500, opened 2024-10-02
Use loops instead of sendgrid for the newsletter subscription

javisperez:
This PR uses Loops instead of Sendgrid for the `subscribe to our newsletter` form in the Hero.

### DIFF
--- a/docs/.vitepress/theme/components/Home.vue
+++ b/docs/.vitepress/theme/components/Home.vue
@@ -86,9 +86,15 @@ const subscribeToNewsletter = async () => {
       }
     })
 
-    await axios.put('https://sendgrid-proxy.gorkem.workers.dev/v3/marketing/contacts', {
-      list_ids: [ LIST_ID ],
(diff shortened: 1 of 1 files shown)


nida-hasan (approved):

docs/.vitepress/theme/components/Home.vue
```diff
@@ -86,9 +86,15 @@ const subscribeToNewsletter = async () => {
...
-      contacts: [ { email: email.value }  ]
+    await axios.post('https://newsprxy.gorkem.workers.dev/', {
+      email: email.value,
+      userGroup: 'KitOps',
+      formName: 'KitOps-Community'
+    }, {
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded',
+        'Expect': ''
+      }
```
gorkem, line 97: This needs to have a more robust error handling. I think you should also check for HTTP 202 response code and fail for 4xx, 5xx

gorkem (changes requested):

PR #539, opened 2024-10-17
Improve the email field validation in the home page

javisperez:
This PR uses a custom (better?) regexp pattern to validate emails in the newsletter subscription form.

### DIFF
--- a/docs/.vitepress/theme/components/Home.vue
+++ b/docs/.vitepress/theme/components/Home.vue
@@ -125,7 +125,7 @@ const subscribeToNewsletter = async () => {
           :disabled="isBusy"
           id="email"
           type="email"
-          pattern="^[a-zA-Z0-9]+(?:\.[a-zA-Z0-9]+)*@[a-zA-Z0-9]+(?:\.[a-zA-Z0-9]+)*$"
+          pattern="^[a-zA-Z0-9]+(?:[._+-][a-zA-Z0-9]+)*@[a-zA-Z0-9-]+\.[a-zA-Z]{2,}$"
           name="email"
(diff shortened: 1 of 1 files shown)


amisevsk (commented):
What, no +emails?

amisevsk (approved):

javisperez:
That my friend @amisevsk was a bug :D nice catch. Updated it to suppot common email characters like `-`, `+`, `.` and `_`

docs/.vitepress/theme/components/Home.vue
```diff
@@ -125,6 +125,7 @@ const subscribeToNewsletter = async () => {
           :disabled="isBusy"
           id="email"
           type="email"
+          pattern="^[a-zA-Z0-9]+(?:\.[a-zA-Z0-9]+)*@[a-zA-Z0-9]+(?:\.[a-zA-Z0-9]+)*$"
```
gorkem: Can we improve it so that we allow for `name+test@mymail.com` kind of mails?

gorkem (changes requested):

docs/.vitepress/theme/components/Home.vue
javisperez: yeah, i just push a fix right before this comment :D it was my mistake, give it another try

javisperez: This is the commit https://github.com/jozu-ai/kitops/pull/539/commits/b550fd01d281bad30f5ca396153cc21547596be0

gorkem (approved):

PR #540, opened 2024-10-17
Fixes a bug in my previous PR.

javisperez:
This PR removes the use of lookahead in the previous pr because we cant use it in input fields...

### DIFF
--- a/docs/.vitepress/theme/components/Home.vue
+++ b/docs/.vitepress/theme/components/Home.vue
@@ -125,7 +125,7 @@ const subscribeToNewsletter = async () => {
           :disabled="isBusy"
           id="email"
           type="email"
-          pattern="^[a-zA-Z0-9]+(?:\.[a-zA-Z0-9]+)*@[a-zA-Z0-9]+(?:\.[a-zA-Z0-9]+)*$"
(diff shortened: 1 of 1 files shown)


docs/.vitepress/theme/components/Home.vue
```diff
@@ -125,7 +125,7 @@ const subscribeToNewsletter = async () => {
           :disabled="isBusy"
           id="email"
           type="email"
-          pattern="^[a-zA-Z0-9]+(?:[._+-][a-zA-Z0-9]+)*@[a-zA-Z0-9-]+\.[a-zA-Z]{2,}$"
+          pattern="^[a-zA-Z0-9]+([._+\-][a-zA-Z0-9]+)*@[a-zA-Z0-9\-]+\.[a-zA-Z]{2,}$"
```
amisevsk, line 128: I don't think you need to escape the `-` -- if it's at the end of the range it's treated as a character (at least in regex parsers I'm familiar with)

amisevsk (approved):

docs/.vitepress/theme/components/Home.vue
javisperez: thats what was my thinking too, but if i dont i get an `invalid range` error :D

javisperez: check how i escaped both dashes, in the user and in the domain. #fun

PR #579, opened 2024-10-30
Docs for containerizing and Kubernetes deployment

bmicklea:
Added new docs and updates to show people how to create containers and Kubernetes deployments directly from ModelKits:
* Added new `deploy.md` page with instructions for init containers, Kube deployments, and custom container creation
* Added references to deployment to existing docs pages
* Added deployment doc to ToC
* Added pointer to issues and Discord to some pages where it was missing

### DIFF
--- a/docs/src/docs/deploy.md
+++ b/docs/src/docs/deploy.md
@@ -17,7 +17,7 @@ The container is configurable via environment variables:
 
 `$MODELKIT_REF`: The ModelKit to pull (required).
 `$UNPACK_PATH`: Where to unpack the ModelKit (normally you’d want a `volumeMount` here). This is required and will default to `/home/user/modelkit`.
-`$UNPACK_FILTER`: Optional filter to limit what is unpacked (e.g., just the model, or model + code). The filter format is the same as the [–filter command line argument](./cli/cli-reference.md) for the Kit CLI.
+`$UNPACK_FILTER`: Optional filter to limit what is unpacked (e.g., just the model, or model + code). The filter format is the same as the [--filter command line argument](./cli/cli-reference.md) for the Kit CLI.
 `$COSIGN_KEY`: Path to the key that should be used for verification, mounted inside the init container (e.g., from a Kubernetes secret).
 `$COSIGN_CERT_IDENTITY`: Signing identity for keyless signing.
 `$COSIGN_CERT_OIDC_ISSUER`: OIDC issuer for keyless signer identity.
@@ -62,7 +62,7 @@ The container is configurable via environment variables:
 
 The containerized Kit CLI can be used to tailor the running of a ModelKit because you can run any Kit CLI command. This gives you flexibility, but more manual work (the world is your oyster, but it may be hard to shuck).
 
-This container runs `kit` as its entrypoint, accepting Kit CLI arguments. You can also replace any time you use a kit CLI command with `docker run ghcr.io/jozu-ai/kit:latest` and it will run Kit inside the container. This means if you want to download it to your local machine, you’ll need to mount a docker volume, but you can use this to inspect ModelKits without installing the CLI.
+This container runs `kit` as its entrypoint, accepting Kit CLI arguments. So you could run the container instead of downloading and installing the Kit CLI, although you’ll need to mount a docker volume.
 
 Docker run example:
 
@@ -109,4 +109,4 @@ Example `dockerfile` for a custom container that has <my-modelkit> built into it
  COPY --from=modelkit-download /tmp/my-modelkit /home/user/modelkit-data
 ```
 
-**Questions or suggestions?** Drop an [issue in our GitHub repository](https://github.com/jozu-ai/kitops/issues) or join [our Discord server](https://discord.gg/3eDb4yAN) to get support or share your feedback.
+**Questions or suggestions?** Drop an [issue in our GitHub repository](https://github.com/jozu-ai/kitops/issues) or join [our Discord server](https://discord.gg/Tapeh8agYy) to get support or share your feedback.
--- a/docs/src/docs/deploy.md
+++ b/docs/src/docs/deploy.md
@@ -1,0 +1,112 @@
+# Deploying ModelKits
+
+This page outlines how to use `init` or Kit CLI containers to deploy a ModelKit-packaged AI/ML project to Kubernetes or any other container runtime. The KitOps repo provides pre-built ModelKits that can be used for both semi-turnkey solutions, and more DIY options.
+
+## Pre-built Containers
+
+There are currently two pre-built containers:
+
+1. Init container: https://github.com/jozu-ai/kitops/blob/main/build/dockerfiles/init/README.md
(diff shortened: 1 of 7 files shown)


docs/src/docs/deploy.md
```diff
@@ -0,0 +1,112 @@
...
+       # Share a volumeMount between the init container and this one
+       volumeMounts:
+         - name: modelkit-storage
+           mountPath: /my-modelkit
+
+   # Run the init container to unpack the ModelKit into the volume mount and make
+   # it available to the main container
+   initContainers:
+     - name: kitops-init
+       image: ghcr.io/jozu-ai/kitops-init-container:latest
```
javisperez, line 45: not sure if it makes sense, but should we use `jozu.ml` instead of `ghcr.io` here?

docs/src/docs/deploy.md
```diff
@@ -0,0 +1,112 @@
...
+     - name: model-server
+       image: "" # Some container that expects your modelkit
+       # Share a volumeMount between the init container and this one
+       volumeMounts:
+         - name: modelkit-storage
+           mountPath: /my-modelkit
+
+   # Run the init container to unpack the ModelKit into the volume mount and make
+   # it available to the main container
+   initContainers:
```
javisperez, line 43: We dont have an `initContainers` entry in the new `inference tab`, should we?

docs/src/docs/deploy.md
```diff
@@ -0,0 +1,112 @@
...
+
+```
+ apiVersion: v1
+ kind: Pod
+ metadata:
+   name: my-modelkit-pod
+ spec:
+   containers:
+     - name: me-using-kit
+       image: ghcr.io/jozu-ai/kit:latest
```
javisperez, line 81: same here with the host (jozu.ml vs ghcr.io)

javisperez (approved):

docs/src/docs/deploy.md
bmicklea: @amisevsk I think you probably chose GHCR because this is actually a container image not a ModelKit - is that right?

docs/src/docs/deploy.md
bmicklea: I think it's the same as the "basic" container. If that's true then I'll update the docs to make that clearer.

docs/src/docs/deploy.md
amisevsk: Yes, for this doc, the container we're talking about is [`ghcr.io/jozu-ai/kitops`](https://github.com/jozu-ai/kitops/pkgs/container/kitops), which is built in this repo.

docs/src/docs/deploy.md
amisevsk: The Kit CLI image is currently only available through ghcr.io and isn't published to jozu.ml (if I'm remembering right).

docs/src/docs/deploy.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Init Container
+
+The init container unpacks the model reference from a ModelKit to a specific path and then exits. This makes it useful as a Kubernetes `init` container. This container also supports verifying signatures for containers automatically from key-based or keyless signers.
+
+The container is configurable via environment variables:
+
+`$MODELKIT_REF`: The ModelKit to pull (required).
+`$UNPACK_PATH`: Where to unpack the ModelKit (normally you’d want a `volumeMount` here). This is required and will default to `/home/user/modelkit`.
+`$UNPACK_FILTER`: Optional filter to limit what is unpacked (e.g., just the model, or model + code). The filter format is the same as the [–filter command line argument](./cli/cli-reference.md) for the Kit CLI.
```
amisevsk, line 20: I think your editor replaced `--` with an em-dash here, which would break copy pasting (if someone happened to do that)
```suggestion
`$UNPACK_FILTER`: Optional filter to limit what is unpacked (e.g., just the model, or model + code). The filter format is the same as the [`--filter` command line argument](./cli/cli-reference.md) for the Kit CLI.
```

docs/src/docs/deploy.md
```diff
@@ -0,0 +1,112 @@
...
+
+   # Define a volume to store the ModelKit
+   volumes:
+     - name: modelkit-storage
+       emptyDir: {}       
+```
+
+## Using the Kit CLI Container
+
+The containerized Kit CLI can be used to tailor the running of a ModelKit because you can run any Kit CLI command. This gives you flexibility, but more manual work (the world is your oyster, but it may be hard to shuck).
```
amisevsk, line 63: Glad this line made it in :wink:

docs/src/docs/deploy.md
```diff
@@ -0,0 +1,112 @@
...
+
+This container runs `kit` as its entrypoint, accepting Kit CLI arguments. You can also replace any time you use a kit CLI command with `docker run ghcr.io/jozu-ai/kit:latest` and it will run Kit inside the container. This means if you want to download it to your local machine, you’ll need to mount a docker volume, but you can use this to inspect ModelKits without installing the CLI.
```
amisevsk, line 65: We can maybe trim out the "replace calling the Kit CLI bit" from this paragraph -- saying "you can run the container instead of the Kit CLI" is a useful description but I worry it will confuse people if they actually try it (e.g. "I ran kit unpack using the docker command and nothing showed up!"

I think I meant this more as a conceptual "think about the container like you think about the CLI" rather than an actual "try it", since the actual try it means getting into docker's semi-arcane volume mounting system.

gorkem (approved):

docs/src/docs/modelkit/intro.md
```diff
@@ -21,3 +21,5 @@ ModelKit revolutionizes the way AI/ML artifacts are shared and managed throughou
 **Optimized for AI/ML Workflows:** ModelKits are tailor-made for AI/ML projects, addressing specific needs such as versioning and environment configuration.
 
 ModelKit is not just a packaging format; it's a building block for innovation, simplifying the complexities of AI/ML development and deployment. By adopting ModelKit, teams can focus more on creating value and less on managing the intricacies of artifact storage and sharing.
+
+**Questions or suggestions?** Drop an [issue in our GitHub repository](https://github.com/jozu-ai/kitops/issues) or join [our Discord server](https://discord.gg/3eDb4yAN) to get support or share your feedback.
```
amisevsk, line 25: A blast from the past, but should we update or remove [this page](https://github.com/jozu-ai/kitops/blob/main/docs/discord-tracking.md)?

amisevsk (approved):

docs/src/docs/deploy.md
bmicklea: @amisevsk please check the updated wording.

docs/src/docs/modelkit/intro.md
bmicklea: OMG, I grabbed the wrong link and then pasted it in many pages...so this PR now updates all those other pages (sigh).

amisevsk (approved):

PR #582, opened 2024-11-01
Update deploy.md

javisperez:
This PR fixes the broken build for the docs.

### DIFF
--- a/docs/src/docs/deploy.md
+++ b/docs/src/docs/deploy.md
@@ -88,14 +88,14 @@ Kubernetes example:
 
 Going a step further you can use the Kit CLI container to create your own bespoke ModelKit containers.
 
-Example `dockerfile` for a custom container that has <my-modelkit> built into it:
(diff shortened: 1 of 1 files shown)


docs/src/docs/deploy.md
```diff
@@ -88,14 +88,14 @@ Kubernetes example:
 
 Going a step further you can use the Kit CLI container to create your own bespoke ModelKit containers.
 
-Example `dockerfile` for a custom container that has <my-modelkit> built into it:
+Example `dockerfile` for a custom container that has `my-modelkit` built into it:
```
javisperez, line 91: if you gonna use `<` and `>` it needs to be wrapped in a inline-code backtick, otherwise, it tries to render it as html and fails.

nida-hasan (approved):

bmicklea:
Ah gotcha, I should have thought of that! Thanks!

PR #610, opened 2024-11-08
[feat:] Add Log tailing

srikary12:
### Description
Adds log tailing capability

### Linked issues
https://github.com/jozu-ai/kitops/issues/591


pkg/lib/harness/llm-harness.go
```diff
@@ -181,8 +183,26 @@ func PrintLogs(configHome string, w io.Writer) error {
 		return fmt.Errorf("error reading log file: %w", err)
 	}
 	defer logFile.Close()
-	if _, err = io.Copy(w, logFile); err != nil {
-		return fmt.Errorf("failed to print log file: %w", err)
+	if !tail {
+		if _, err = io.Copy(w, logFile); err != nil {
+			return fmt.Errorf("failed to print log file: %w", err)
+		}
```
gorkem, line 189: You can simplify early returning in here.  Alternatively you can handle the `!tail` case in the loop.
```
return nil
```

pkg/lib/harness/llm-harness.go
```diff
@@ -181,8 +183,26 @@ func PrintLogs(configHome string, w io.Writer) error {
...
+			if err != nil {
+				if err == io.EOF {
+					time.Sleep(1 * time.Second)
+				} else {
+					return fmt.Errorf("failed to print log file: %w", err)
+				}
+			}
+			stringReader := strings.NewReader(line)
+			if _, err = io.Copy(w, stringReader); err != nil {
+				return fmt.Errorf("failed to print log file: %w", err)
```
gorkem, line 203: This is unncessary overhead you can write the line here instead
```suggestion
	        if _, err := w.Write([]byte(line)); err != nil {
                return fmt.Errorf("failed to write to output: %w", err) 
```

gorkem (changes requested):
This is not handling the case when the server is stopped and started. I expect the log following to either drop or switch to the newly started server instance.

srikary12:
I can drop the logs when server is stopped.

Do we cuurently allow 2 development servers running at once?

gorkem:
> I can drop the logs when server is stopped.
> 
> Do we currently allow 2 development servers running at once?

Only one instance is allowed to run.

srikary12:
@gorkem shows a server stopped message when the server is stopped and resolves conversations :).

pkg/lib/harness/llm-harness.go
```diff
@@ -181,10 +183,35 @@ func PrintLogs(configHome string, w io.Writer) error {
...
+			}
+		}
+		if _, err := w.Write([]byte(line)); err != nil {
+			return fmt.Errorf("failed to write to output: %w", err)
+		}
 	}
-	return nil
+}
+
+func checkExistance(configHome string) bool {
```
gorkem, line 210: Can you fix the spelling? It should be `checkExistence`

gorkem (changes requested):
Realized my inline comments are not uploaded to GH trying again

pkg/lib/harness/llm-harness.go
```diff
@@ -181,10 +183,35 @@ func PrintLogs(configHome string, w io.Writer) error {
...
 	}
-	return nil
+}
+
+func checkExistence(configHome string) bool {
+	pidFile := filepath.Join(constants.HarnessPath(configHome), constants.HarnessProcessFile)
+
+	_, err := readPIDFromFile(pidFile)
+	return os.IsNotExist(err)
 }
```
amisevsk, line 215: This is just checking if the PID file exists, not whether the server is actively running. I think we also need to use `isProcessRunning(pid)`

srikary12: When we run `kit dev stop` we are verifying the the running server this way, and stopping it. Used the same logic in here.

srikary12: `func (harness *LLMHarness) Stop() error {
	pidFile := filepath.Join(constants.HarnessPath(harness.ConfigHome), constants.HarnessProcessFile)

	pid, err := readPIDFromFile(pidFile)
	if os.IsNotExist(err) {
		return fmt.Errorf("no Running server found")
	}
	if err != nil {
		return err
	}
`

This is how we are checking the server status in stop interface.

pkg/lib/harness/llm-harness.go
```diff
@@ -181,10 +183,35 @@ func PrintLogs(configHome string, w io.Writer) error {
...
+			if err == io.EOF {
+				if !follow {
+					return nil
+				}
+				time.Sleep(1 * time.Second)
+				serverStatus := checkExistence(configHome)
+				if serverStatus {
+					return fmt.Errorf("server stopped")
+				}
+				continue
```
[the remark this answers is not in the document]
srikary12: Will resolve this

srikary12:
@amisevsk We still need to show logs even if server crashes. I think including the above case wouldn't display logs when the server crashes. We only remove pid file once server ends properly until then we display logs.

pkg/lib/harness/llm-harness.go
srikary12: We still need to show logs even if server crashes. I think including the above case wouldn't display logs when the server crashes. We only remove pid file once server ends properly until then we display logs.

amisevsk: These checks run only when reaching EOF on the file though -- we've already printed all existing logs at this point. If the server has crashed, there aren't going to be any more logs to print (same as the server terminated normally case)

srikary12: Okay, will add that case too.

Issue #626, opened 2024-11-26
CodeQL Finding: Writable file handle closed without error handling - file: llm_payload_all.go


tlgevers:
KitOps is an integral part of our MLOps processes & subject to our internal review and scans.

The following was raised in a recent report generated with CodeQL:

Severity: High
> File handle may be writable as a result of data flow from a [call to OpenFile](1) and closing it may result in data loss upon failure, which is not handled explicitly

Found within:
pkg/lib/harness/llm_payload_all.go
Lines 65 & 105

PR #628, opened 2024-11-26
Handle file close in defer better for dev command

### LINKED ISSUE #625: CodeQL Finding: Writable file handle closed without error handling - file: llm-harness.go
KitOps is an integral part of our MLOps processes & subject to our internal review and scans.

The following was raised in a recent report generated with CodeQL:

Severity: High
> File handle may be writable as a result of data flow from a [call to OpenFile](1) and closing it may result in data loss upon failure, which is not handled explicitly

Found within:
pkg/lib/harness/llm-harness.go
Line 111

gorkem:
Improve the handling of file close in defer by checking for errors during the close

fixes #625, #626

### DIFF
--- a/pkg/lib/harness/llm_payload_all.go
+++ b/pkg/lib/harness/llm_payload_all.go
@@ -29,7 +29,7 @@ import (
 	"kitops/pkg/lib/filesystem"
 )
 
-func extractFile(fs fs.FS, file, harnessHome string) error {
+func extractFile(fs fs.FS, file, harnessHome string) (err error) {
 	srcFile, err := fs.Open(file)
 	if err != nil {
 		return fmt.Errorf("failed to open file %s: %w", file, err)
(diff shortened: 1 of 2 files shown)


pkg/lib/harness/llm_payload_all.go
```diff
@@ -118,3 +118,15 @@ func extractTar(tr *tar.Reader, dir string) error {
...
+	defer func() {
+		if cerr := f.Close(); cerr != nil {
+			if *err == nil {
+				*err = fmt.Errorf("%s: %w", msg, cerr)
+			} else {
+				*err = fmt.Errorf("%v; %s: %w", *err, msg, cerr)
+			}
+		}
+	}()
+}
```
amisevsk, line 132: I don't think this will work; the defer will execute when leaving the current scope, which is the called function (`deferCloseFile`)

amisevsk (approved):

PR #629, opened 2024-11-27
Updates to dev UI

javisperez:
This PR updates the Dev Mode UI to a new look. With changes in the UX, the flow of the data and useful snippets for Python, Node and SH.


frontend/dev-mode/index.html
amisevsk, line 1: Not related to this file, but the `.nvmrc` in `frontend/dev-mode` is still on v18.

... also I wasn't able to build it locally using v20 or v22 (`pnpm i` was failing)

amisevsk (commented):
Is there a good way to test this locally with a running `kit dev` server? I tried `pnpm dev` but the conversation hangs, and `kit dev` is still serving the old one.

amisevsk:
On firefox, it looks like there's a discrepancy in text size between the code block on the right and the text on the left (14px vs 18px), which looks a little weird to me:

image

C.f. 14px code with a 1.5 line height (I just edited the page locally):

image

gorkem:
This is what I got when I tried it. 
I have built the UI with 
```
go generate ./...
```
Are there any build changes that we should reflect to the generate script?


Screenshot 2024-11-27 at 5 49 43 PM

frontend/dev-mode/index.html
javisperez: ill update the nvmrc thanks for the catch. About running it locally, the way i do it to test it is that i run kit unpack to get a kitfile + model and then do:
```
kit dev start ./unpacked --port 64246
```

and then in the `frontend/dev-mode/src/services/completion.ts` file i update the `apiUrl` const and point it to `http://localhost:64246` (change that i dont commit btw), and

```
pnpm dev
```

javisperez:
> This is what I got when I tried it. I have built the UI with
> 
> ```
> go generate ./...
> ```
> 
> Are there any build changes that we should reflect to the generate script?
> 
> Screenshot 2024-11-27 at 5 49 43 PM

@gorkem i dont understand that question :D what is the `generate script` ? is that for me or Angel?

also, why are your scrollbars so weird looking? that looks off... ill try to reproduce..

gorkem:
@javisperez when we build and package the UI as part of the package it uses https://github.com/javisperez/kitops/blob/8b0364757e5e18dd329022f08e43c9aa38a12605/pkg/lib/harness/generated/gen_common.sh#L6-L15 Can you check if this looks correct?

javisperez:
@gorkem @amisevsk things should be way better now, feel free to pull and try again please :D.

amisevsk:
I totally forgot about go generate :D 

We should add it to the frontend readme

gorkem:
The build issues are gone for me... But I still see the ugly scrollbars on the code section. It feels like the font for the code is disproportionately larger with the rest of the UI.

Also, should the code be highlighted? 

image

amisevsk (approved):
Tested on firefox and chrome (Linux) and it looks good to me. I can't reproduce the scrollbars issue Gorkem mentioned.

The "Response" panel on the landing page doesn't really make sense to me though:

image

javisperez:
@amisevsk that looks like a bug :flip-the-table: on it

javisperez:
@gorkem @amisevsk wanna give it another try? i cant reproduce the scrollbar issues but i did fix a few things with the highlight js dependency, hopefully that will fix it.

amisevsk (approved):
Tested latest changes with `pnpm preview` and the syntax highlighting is working. On Linux, I'm seeing normal (small) scroll bars.

javisperez:
@gorkem @amisevsk added custom scrollbar styles that should be consistent across devices (trackpad vs mouse) feel free to try again :D

frontend/dev-mode/src/components/InputCodeSnippets.vue
```diff
@@ -0,0 +1,173 @@
...
+    .filter(Boolean)
+    .reverse()
+    .join(',')
+    }
+  ],
+  "model": "${currentModel.value}",
+  "stream": true,
+  ${
+    Object.entries(parameters.value).map(([key, value]) => {
+      return `"${key}"": ${value}`
```
gorkem, line 114: There is an extra `"` that is breaking the snippets. You can see that the json on shell Tab is malformed

.github/workflows/build-devmode.yaml
```diff
@@ -0,0 +1,56 @@
...
+env:
+  PNPM_CACHE_STORE_PATH: ""
+
+jobs:
+  build-ui:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
gorkem, line 21: These are older versions of the actions... We also pin the actions to their SHA values and let dependabot do the updates. See platform-relese workflow which uses the same actions for the current versions in use.

frontend/dev-mode/src/components/ChatPrompt.vue
```diff
@@ -0,0 +1,237 @@
...
+const isSettingsModalOpen = ref(false)
+
+const parameters = inject<RemovableRef<Parameters>>('parameters', ref(DEFAULT_PARAMS_VALUES))
+
+const session = inject<Ref<Session>>('session', ref(DEFAULT_SESSION))
+const stats = inject('stats', {} as Stats)
+const isGenerating = inject('isGenerating', false)
+const isPending = inject('isPending', false)
+const isChatStarted = inject('isChatStarted', false)
+const runChat = inject('runChat', (message: string) => message)
```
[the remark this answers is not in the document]
javisperez: these fallback functions are only here to make TS happy, this will never happen because these injects are passed by the parent high order component, which means that the app will break if the provided functions / values are not available, but ill add these exits just get this PR pushed :)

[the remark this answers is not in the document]
javisperez: also, the typing cant be `runChat` so it'll be slightly different than your suggestion, it'll be a generic function

frontend/dev-mode/src/components/ui/CodeHighlighter.vue
```diff
@@ -0,0 +1,67 @@
...
+const highlightCode = async () => {
+  if (!codeRef.value) {
+    return
+  }
+
+  codeRef.value.textContent = props.code
+
+  await nextTick()
+
+  if (codeRef.value) {
```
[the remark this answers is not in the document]
javisperez: We do, the markdown library seems to do some sort of "flickering" after the change in line 26 of this file (`codeRef.value.textContent = props.code`) which causes codeRef.value to be null (i assume the element gets recreated?) for a brief time. I used to have a comment about it here, i dont rememebr removing it 😬

gorkem (approved):

PR #635, opened 2024-12-03
Updates to ModelKit format

### LINKED ISSUE #489: Consider QoL improvements to tar artifacts generated by kit
When packing a modelkit, each layer consists of a tar of the files in that layer's directory. For example, packing a model layer
```yaml
model:
  path: model-files/
```
will result in a plain tar containing the contents of `model-files` (but not the directory itself). During unpacking, this context is reconstructed from the Kitfile so that digests are unchanged.

In addition, creating reproducible modelkits (unpacking and repacking results in the same digest) requires some additional changes, such as setting owner/group, timestamps, etc. to a known value.

This means that the tar files generated by Kit lose some information from the packing environment -- tooling using modelkits depends on kit-specific logic for extracting the contents of a modelkit.

To make layer tars more useful outside of a kit context, it would be useful to find a way to include more of the context from the packing environment (e.g. directory names).

**Additional context**
Considering this change, it's likely something we want to decide sooner rather than later; any change to the pack process will break reproducibility of repacking unpacked modelkits. We also need to consider compatibility, so that previously-packed modelkits can still be unpacked with whatever new format we choose.

amisevsk:
### Description
This PR implements the updates discussed in #489:

* We've added fields to the ModelKit manifest's config to store layer digests and diff IDs, so that we can match Kitfile fields to layers more easily (rather than relying on order)
* We're now packing tar layers slightly differently -- instead of packing the contents of a layer's path, we pack full directory structure from the pack context. In short, if you extract the tarballs for all layers in a manifest, you should get back the original file structure

This change is backwards compatible with ModelKits packed prior to this change -- they can still be unpacked, etc. This change does, however, break reproducible digests for older ModelKits (packing the same data with Kit prior to this change will result in a ModelKit with a different digest due to the changes to tar files and the config). Within Kit versions, digests are still reproducible.  

### Linked issues
Closes #489

### DIFF
--- a/pkg/cmd/unpack/unpack.go
+++ b/pkg/cmd/unpack/unpack.go
@@ -77,20 +77,33 @@ func runUnpackRecursive(ctx context.Context, opts *unpackOptions, visitedRefs []
 
 	// Since there might be multiple datasets, etc. we need to synchronously iterate
 	// through the config's relevant field to get the correct path for unpacking
+	// We need to support older ModelKits (that were packed without diffIDs and digest
+	// in the config) for now, so we need to continue using the old structure.
 	var modelPartIdx, codeIdx, datasetIdx, docsIdx int
 	for _, layerDesc := range manifest.Layers {
+		// This variable supports older-format tar layers (that don't include the
+		// layer path). For current ModelKits, this will be empty
 		var relPath string
+
 		mediaType := constants.ParseMediaType(layerDesc.MediaType)
 		switch mediaType.BaseType {
 		case constants.ModelType:
 			if !shouldUnpackLayer(config.Model, opts.filterConfs) {
 				continue
 			}
-			_, relPath, err = filesystem.VerifySubpath(opts.unpackDir, config.Model.Path)
-			if err != nil {
-				return fmt.Errorf("error resolving model path: %w", err)
+			if config.Model.LayerInfo != nil {
+				if config.Model.LayerInfo.Digest != layerDesc.Digest.String() {
+					return fmt.Errorf("digest in config and manifest do not match in model")
+				}
+				relPath = ""
+			} else {
+				_, relPath, err = filesystem.VerifySubpath(opts.unpackDir, config.Model.Path)
+				if err != nil {
+					return fmt.Errorf("error resolving model path: %w", err)
+				}
 			}
-			output.Infof("Unpacking model %s to %s", config.Model.Name, relPath)
+
+			output.Infof("Unpacking model %s to %s", config.Model.Name, config.Model.Path)
 
 		case constants.ModelPartType:
 			part := config.Model.Parts[modelPartIdx]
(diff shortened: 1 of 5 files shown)


pkg/cmd/unpack/unpack.go
```diff
@@ -77,32 +77,52 @@ func runUnpackRecursive(ctx context.Context, opts *unpackOptions, visitedRefs []
...
 		mediaType := constants.ParseMediaType(layerDesc.MediaType)
 		switch mediaType.BaseType {
 		case constants.ModelType:
 			if !shouldUnpackLayer(config.Model, opts.filterConfs) {
 				continue
 			}
-			_, relPath, err = filesystem.VerifySubpath(opts.unpackDir, config.Model.Path)
-			if err != nil {
-				return fmt.Errorf("error resolving model path: %w", err)
+			if config.Model.LayerInfo != nil {
```
gorkem, line 94: Should this also check if there is a `config.Model` ?

pkg/cmd/unpack/unpack.go
```diff
@@ -137,9 +171,16 @@ func runUnpackRecursive(ctx context.Context, opts *unpackOptions, visitedRefs []
...
+			if docsEntry.LayerInfo != nil {
+				if docsEntry.LayerInfo.Digest != layerDesc.Digest.String() {
+					return fmt.Errorf("digest in config and manifest do not match in docs layer")
+				}
+				relPath = ""
+			} else {
+				_, relPath, err = filesystem.VerifySubpath(opts.unpackDir, docsEntry.Path)
+				if err != nil {
+					return fmt.Errorf("error resolving path %s for docs: %w", docsEntry.Path, err)
+				}
```
gorkem, line 183: Duplicated code for handling every layer. Should it be refactored ?

gorkem (approved):

pkg/lib/kitfile/layer.go
```diff
@@ -136,32 +119,48 @@ func compressLayer(path string, mediaType constants.MediaType, ignore filesystem
...
+		bToA, errB := filepath.Rel(b, a)
+		if errA != nil || errB != nil {
+			plog.Logf(output.LogLevelWarn, "Cannot compare directories %s and %s, skipping path", a, b)
+			return false
+		}
+		if strings.Contains(aToB, "..") && strings.Contains(bToA, "..") {
+			return false
+		}
+		return true
 	}
```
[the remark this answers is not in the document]
amisevsk: It might be, I opted for Contains so as to not worry about `./..`, `C:path`, etc.

I'd have to test how relative paths are handled on Windows to be sure.

pkg/cmd/unpack/unpack.go
amisevsk: We only enter this branch of the switch if we encounter a model layer, which implies the existence of a `model` field in the Kitfile (i.e. how did we pack a model layer without a model section in the Kitfile?).

pkg/cmd/unpack/unpack.go
amisevsk: This is just kind of the way things end up happening in Golang. I can try to rework it but in my experience it ends up decreasing readability :)

amisevsk: Latest commit should make this a little clearer.

PR #638, opened 2024-12-04
Add "Subscribe to KitOps" section

### LINKED ISSUE #632: UI: Add the newsletter form to KitOps
We want to add the subscription form for the newsletter in the KitOps site again, but this time, we want to try using CF's captcha instead of recaptcha v3.

javisperez:
Fixes #632 by adding a "Subscribe to KitOps" section in the home page (previously a form in the Hero section).

This also uses CF Turnstile instead of reCaptcha v3.

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -129,7 +129,7 @@ export default defineConfig({
     ],
 
     footer: {
-      message: '<a href="https://opensource.org/licenses/MIT" target="_blank">MIT License</a>',
+      message: 'Made with <3 by Jozu',
       copyright: `Copyright © ${new Date().getFullYear()} Jozu`
     }
   },
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -24,8 +24,6 @@ const head = [
   ['script', {}, "window.dataLayer = window.dataLayer || [];\nfunction gtag(){dataLayer.push(arguments);}\ngtag('js', new Date());\ngtag('config', 'G-QTDTMG01Z5');"],
   ['script', {}, "(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':new Date().getTime(),event:'gtm.js'});\nvar f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';\nj.async=true;j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);\n})(window,document,'script','dataLayer','GTM-TFFZXCQW');"],
   ['script', { async: '', defer: '', src: 'https://buttons.github.io/buttons.js' }],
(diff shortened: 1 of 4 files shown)


docs/.vitepress/config.mts
javisperez, line 1: all the changes in this file are just clean up from moving to a custom sitemap script to the official sitemap setting.

docs/.vitepress/theme/components/Home.vue
```diff
@@ -4,11 +4,19 @@ import { ref, computed } from 'vue'
...
 const error = ref('')
 const email = ref('')
+const token = ref('')
 const favoriteDevOpsTool = ref('')
 const isBusy = ref(false)
+const isSubscribed = ref(false)
+const isSuccess = ref(false)
+const isProd = import.meta.env.PROD
+
+isSubscribed.value = localStorage.getItem('subscribed') === 'true'
```
javisperez, line 19: yes, a string, localStorage doesn't handle boolean values and TS knows it lol.

docs/.vitepress/theme/components/Home.vue
```diff
@@ -93,6 +126,50 @@ const quotesOffsetDesktop = computed(() => `translateX(${((activeQuote.value * 6
   </video>
 </div>
 
+<div v-if="!isSubscribed" class="mt-32 md:mt-40 xl:mt-60 px-6 md:px-12 content-container">
+  <h2 class="text-center">stay informed About Kitops</h2>
```
gorkem, line 130: stay => Stay. It should start with uppercase

docs/.vitepress/config.mts
```diff
@@ -131,11 +127,9 @@ export default defineConfig({
         link: 'https://discord.gg/Tapeh8agYy'
       },
     ],
+
     footer: {
-      license: {
-        text: 'MIT License',
-        link: 'https://opensource.org/licenses/MIT'
-      },
+      message: '<a href="https://opensource.org/licenses/MIT" target="_blank">MIT License</a>',
```
gorkem, line 132: What is this license related to ? The license for kitops is Apache 2.0

gorkem (approved):
One typo to be fixed before merge

docs/.vitepress/theme/components/Home.vue
javisperez: its not a typo :D its a decision. The font renders uppercase and lowercase completely different, so we make some "play" with those to get the style that we want. Example: https://github.com/jozu-ai/kitops/blob/3c9e7aa29ebe399609b6df4488849a2faeaf9585/docs/.vitepress/theme/components/Home.vue#L200

And if i make that text capitalized, it looks like this, which is no bueno:

image

docs/.vitepress/config.mts
javisperez: i can update it but this is not really used. We have a custom Vue component for the footer (only used in the homepage), but if we don't add this entry to the config, it wont render it _at all_. Its a Vitepress thing sadly...But ill use apache 2 just in case.

javisperez: nvm, i just updated the text to not mention any license

PR #645, opened 2024-12-05
Add Model format field to Kitfile

amisevsk:
### Description
To make runtime usage simpler, add Kitfile field .model.format. This can be set to store the model's data format (e.g. safetensors, gguf, onnx, etc.).

This information can be used by runtimes to determine how to consume the model (e.g. if the runtime can handle .gguf files)

### Linked issues
No issue at the moment


gorkem:
Do we also need to record the `format version` some formats like onnx change quite often?

amisevsk:
Maybe, but adding that is a backwards-compatible change we can make if we decide we need it. I'd rather not throw fields into the spec unless we know they're going to be used somewhere.

gorkem:
the reason I brought this up on the annotations that we had proposed on model specification group, I had it listed as `format.name` and `format.version`.

PR #646, opened 2024-12-05
Add Discord popup modal

javisperez:
This PR adds a new modal in the `/docs` section only that shows a link to join our Discord server.

### DIFF
--- a/docs/.vitepress/theme/components/DiscordModal.vue
+++ b/docs/.vitepress/theme/components/DiscordModal.vue
@@ -50,7 +50,7 @@ onBeforeUnmount(() => {
       </h3>
 
       <div class="flex justify-center items-center">
-        <a href="https://discord.gg/Tapeh8agYy" class="kit-button kit-button-gold text-center mt-10 mx-auto inline-block w-max">
+        <a href="https://discord.gg/Tapeh8agYy" target="_blank" class="kit-button kit-button-gold text-center mt-10 mx-auto inline-block w-max">
           Say hello
         </a>
       </div>
--- a/docs/.vitepress/theme/components/DiscordModal.vue
+++ b/docs/.vitepress/theme/components/DiscordModal.vue
@@ -1,0 +1,70 @@
+<script setup lang="ts">
+import { onBeforeUnmount, onMounted, ref } from 'vue'
+
+const ONE_DAY_MS = 1000 * 60 * 60 * 24
+const STORAGE_KEY = 'discord-modal-last-opened'
+
+const isVisible = ref(false)
+
+const close = () => {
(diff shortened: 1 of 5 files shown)


docs/.vitepress/theme/components/DiscordModal.vue
```diff
@@ -0,0 +1,70 @@
...
+
+const close = () => {
+  isVisible.value  = false
+  localStorage.setItem(STORAGE_KEY, String(new Date().getTime()))
+}
+
+onMounted(() => {
+  // get the last time we showed the modal, or 1 day ago if it's not set yet
+  const lastOpenedAt = localStorage.getItem(STORAGE_KEY) || (new Date().getTime()) - ONE_DAY_MS
+  const lastOpenDiff = new Date().getTime() - new Date(Number(lastOpenedAt)).getTime()
```
gorkem, line 17: Quite a few `Date` object created. Perhaps, it is possible to achieve the same without this many.

docs/.vitepress/theme/components/DiscordModal.vue
```diff
@@ -0,0 +1,70 @@
...
+  if (lastOpenDiff >= ONE_DAY_MS) {
+    // Show the modal 5 seconds after the page load
+    setTimeout(() => {
+      isVisible.value = true
+    }, 5000)
+  }
+})
+
+onBeforeUnmount(() => {
+  close()
```
gorkem, line 29: Should this check if the moda is visible?

gorkem (approved):

docs/.vitepress/theme/components/DiscordModal.vue
javisperez: Sure can do :cattyping:

docs/.vitepress/theme/components/DiscordModal.vue
javisperez: we can, but there's no actual benefit for that because this component only renders (yet it mounts) if the `isVisible` flag is true, (check line 35) so even if we dont check it, setting it false will remove the element from the dom tree and its bindings. We can check it inside the close function too but will be like having two `if` conditions of the same `isVisible` flag and with the same result.

docs/.vitepress/theme/components/Home.vue
```diff
@@ -93,6 +91,8 @@ const subscribeToNewsletter = async () => {
 }
 
 onMounted(() => {
+  isSubscribed.value = localStorage.getItem('subscribed') === 'true'
```
javisperez, line 94: this was generating a warning when building the server part.

PR #661, opened 2024-12-12
Add ability to automatically generate Kitfiles

amisevsk:
### Description
Add `kit init` command, that can be used to generate a basic Kitfile from the contents of a directory. It basically does this by looking at files and trying to guess what they might be, based on file extensions.

I've tested it against a few hand-written Kitfiles and (modulo user-provided metadata) it does an okay job :)

### Linked issues
Closes https://github.com/jozu-ai/kitops/issues/124
Related: https://github.com/jozu-ai/kitops/issues/564

### DIFF
--- a/pkg/cmd/init/cmd.go
+++ b/pkg/cmd/init/cmd.go
@@ -18,7 +18,9 @@ package kitinit
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io/fs"
 	"kitops/pkg/artifact"
 	"kitops/pkg/lib/constants"
 	"kitops/pkg/lib/kitfile"
@@ -41,7 +43,10 @@ By default the command will prompt for input for a name and description for the 
 kit init .
 
 # Generate a Kitfile for files in ./my-model, with name "mymodel" and a description:
-kit init ./my-model --name "mymodel" --desc "This is my model's description"`
+kit init ./my-model --name "mymodel" --desc "This is my model's description"
+
+# Generate a Kitfile, overwriting any existing Kitfile:
+kit init ./my-model --force`
 )
 
 type initOptions struct {
@@ -49,6 +54,7 @@ type initOptions struct {
 	configHome          string
 	modelkitName        string
 	modelkitDescription string
+	overwrite           bool
 }
 
 func InitCommand() *cobra.Command {
@@ -65,6 +71,7 @@ func InitCommand() *cobra.Command {
 
 	cmd.Flags().StringVar(&opts.modelkitName, "name", "", "Name for the ModelKit")
 	cmd.Flags().StringVar(&opts.modelkitDescription, "desc", "", "Description for the ModelKit")
+	cmd.Flags().BoolVarP(&opts.overwrite, "force", "f", false, "Overwrite existing Kitfile if present")
 	cmd.Flags().SortFlags = false
 	return cmd
 }
@@ -83,6 +90,16 @@ func runCommand(opts *initOptions) func(*cobra.Command, []string) error {
 			}
 		}
 
+		// Check for existing Kitfile
+		kitfilePath := filepath.Join(opts.path, constants.DefaultKitfileName)
+		if _, err := os.Stat(kitfilePath); err == nil {
+			if !opts.overwrite {
+				return output.Fatalf("Kitfile already exists at %s. Use '--force' to overwrite", kitfilePath)
+			}
+		} else if !errors.Is(err, fs.ErrNotExist) {
+			return output.Fatalf("Error checking for existing Kitfile: %s", err)
+		}
+
 		kitfile, err := kitfile.GenerateKitfile(opts.path, modelPackage)
 		if err != nil {
 			return output.Fatalf("Error generating Kitfile: %s", err)
@@ -91,7 +108,6 @@ func runCommand(opts *initOptions) func(*cobra.Command, []string) error {
 		if err != nil {
 			return output.Fatalf("Error formatting Kitfile: %s", err)
 		}
-		kitfilePath := filepath.Join(opts.path, constants.DefaultKitfileName)
 		if err := os.WriteFile(kitfilePath, bytes, 0644); err != nil {
 			return output.Fatalf("Failed to write Kitfile: %s", err)
 		}
--- a/pkg/cmd/init/cmd.go
+++ b/pkg/cmd/init/cmd.go
@@ -19,8 +19,10 @@ package kitinit
 import (
 	"context"
 	"fmt"
+	"kitops/pkg/artifact"
 	"kitops/pkg/lib/constants"
 	"kitops/pkg/lib/kitfile"
+	"kitops/pkg/lib/util"
 	"kitops/pkg/output"
 	"os"
 	"path/filepath"
@@ -43,8 +45,10 @@ kit init ./my-model --name "mymodel" --desc "This is my model's description"`
 )
 
 type initOptions struct {
-	path       string
-	configHome string
+	path                string
+	configHome          string
+	modelkitName        string
+	modelkitDescription string
 }
 
 func InitCommand() *cobra.Command {
@@ -59,6 +63,9 @@ func InitCommand() *cobra.Command {
 		Args:    cobra.ExactArgs(1),
 	}
 
+	cmd.Flags().StringVar(&opts.modelkitName, "name", "", "Name for the ModelKit")
+	cmd.Flags().StringVar(&opts.modelkitDescription, "desc", "", "Description for the ModelKit")
+	cmd.Flags().SortFlags = false
 	return cmd
 }
 
@@ -68,7 +75,15 @@ func runCommand(opts *initOptions) func(*cobra.Command, []string) error {
 			return output.Fatalf("Invalid arguments: %s", err)
 		}
 
-		kitfile, err := kitfile.GenerateKitfile(opts.path, nil)
+		var modelPackage *artifact.Package
+		if opts.modelkitName != "" || opts.modelkitDescription != "" {
+			modelPackage = &artifact.Package{
+				Name:        opts.modelkitName,
+				Description: opts.modelkitDescription,
+			}
+		}
+
+		kitfile, err := kitfile.GenerateKitfile(opts.path, modelPackage)
 		if err != nil {
 			return output.Fatalf("Error generating Kitfile: %s", err)
 		}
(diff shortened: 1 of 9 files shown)


pkg/lib/kitfile/generate.go
```diff
@@ -0,0 +1,274 @@
...
+
+var docsSuffixes = []string{
+	".md", ".adoc", ".html", ".pdf",
+}
+
+var metadataSuffixes = []string{
+	".json", ".yaml", ".xml", ".csv", ".txt",
+}
+
+var datasetSuffixes = []string{
```
gorkem, line 53: How about `.parquet` and `.csv`?

pkg/lib/kitfile/generate.go
```diff
@@ -0,0 +1,274 @@
...
+	"kitops/pkg/artifact"
+	"kitops/pkg/output"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/google/licensecheck"
+)
+
+var modelWeightsSuffixes = []string{
```
gorkem, line 31: Missing `.joblib`

pkg/cmd/init/cmd.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (opts *initOptions) complete(ctx context.Context, args []string) error {
+	configHome, ok := ctx.Value(constants.ConfigKey{}).(string)
+	if !ok {
+		return fmt.Errorf("default config path not set on command context")
+	}
+	opts.configHome = configHome
+	opts.path = args[0]
+
+	if opts.modelkitName == "" {
```
gorkem, line 112: Should we still prompt for if name or description has been passed as options?

pkg/lib/kitfile/generate.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+var datasetSuffixes = []string{
+	".tar", ".zip",
+}
+
+// Generate a basic Kitfile by looking at the contents of a directory. Parameter
+// packageOpt can be used to define metadata for the Kitfile (i.e. the package
+// section), which is left empty if the parameter is nil.
+func GenerateKitfile(baseDir string, packageOpt *artifact.Package) (*artifact.KitFile, error) {
```
gorkem, line 60: I think this needs to look into the sub directories as well. If somebody organizes their files under `data/` for instance `data/training.csv` it should be detected.

pkg/lib/kitfile/generate.go
```diff
@@ -0,0 +1,274 @@
...
+		kitfile.Code[0].License = detectedLicenseType
+	} else {
+		kitfile.Package.License = detectedLicenseType
+	}
+
+	return kitfile, nil
+}
+
+func addDirToKitfile(kitfile *artifact.KitFile, path string, d fs.DirEntry) error {
+	// TODO: consider looking into directories to see if we can figure out what they store? Might work for datasets
```
gorkem, line 176: +1 works especially for multimodal.

gorkem (changes requested):

gorkem:
It may be worth linking to this list here. https://github.com/trailofbits/ml-file-formats?tab=readme-ov-file

pkg/lib/util/input.go
```diff
@@ -0,0 +1,50 @@
...
+		bytes, err = term.ReadPassword(int(syscall.Stdin))
+		fmt.Print("\n")
+	} else {
+		reader := bufio.NewReader(os.Stdin)
+		bytes, err = reader.ReadBytes('\n')
+	}
+	if err != nil {
+		return "", fmt.Errorf("failed to read input: %w", err)
+	}
+	return strings.TrimSpace(string(bytes)), nil
```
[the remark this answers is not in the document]
amisevsk: strings.TrimSpace only trims leading and trailing whitespace.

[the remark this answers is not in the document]
gorkem: right... but I was thinking is there a case where we want to keep the leading/trailing spaces

pkg/cmd/init/cmd.go
```diff
@@ -0,0 +1,128 @@
...
+
+		var modelPackage *artifact.Package
+		if opts.modelkitName != "" || opts.modelkitDescription != "" {
+			modelPackage = &artifact.Package{
+				Name:        opts.modelkitName,
+				Description: opts.modelkitDescription,
+			}
+		}
+
+		kitfile, err := kitfile.GenerateKitfile(opts.path, modelPackage)
```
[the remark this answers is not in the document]
amisevsk: Good catch! I'll add a check + a `--force` flag.

pkg/lib/kitfile/generate.go
amisevsk: I'm including `.csv` in metadata, though maybe it makes more sense as a dataset exclusively.

pkg/cmd/init/cmd.go
amisevsk: We'll only prompt if we didn't get a name/description (and if the session is interactive). If you specify `--name` or `--desc` we don't prompt (unless you somehow pass in an empty name/description).

pkg/lib/kitfile/generate.go
amisevsk: Yeah, I was hoping to defer that to a future improvement since it introduces a fair bit of complexity in the general case.

* If we're trying to grab specific files (e.g. a dataset layer for `path: data/training.csv`, then we need to decide what to do with the rest of the directory
  * If that's the only file, we don't need to include the directory itself, since it'll be included in the dataset layer, which means signalling another condition back to the caller
  * If there are other files, we need to either classify them or include the whole directory as code (?) layer
  * Then we also need to handle the case where we've got a catch-all `code` section
* If we want to include the whole directory (i.e. `path: data/`) then we need to figure out a heuristic for deciding what type the whole directory is (most common file type? I'm not sure here)
* If a `data/` dir has 20 .csv files, do we include a single `data/` layer, or generate a Kitfile with 20 dataset layers (one for each file).

amisevsk:
> It may be worth linking to this list here. https://github.com/trailofbits/ml-file-formats?tab=readme-ov-file

I glanced over that list at one point, but I'm not sure how valuable it is in general. I can add it but it doesn't feel like an authoritative source for this sort of thing.

pkg/lib/kitfile/generate.go
amisevsk: Alright, I've added a (messy) implementation of this in commit adf3372. For e.g. https://huggingface.co/onnx-community/dpt-dinov2-small-kitti/tree/main it generates the following Kitfile:

```yaml
manifestVersion: 1.0.0
model:
  name: onnx/model
  path: onnx/model.onnx
  parts:
    - path: onnx/model_bnb4.onnx
    - path: onnx/model_fp16.onnx
    - path: onnx/model_int8.onnx
    - path: onnx/model_q4.onnx
    - path: onnx/model_q4f16.onnx
    - path: onnx/model_quantized.onnx
    - path: onnx/model_uint8.onnx
    - path: config.json
    - path: preprocessor_config.json
    - path: quantize_config.json
code:
  - path: .
docs:
  - path: README.md
    description: Readme file
```
where the `code` section includes the `.gitattributes` file and the `.git` directory (since I cloned it locally)

pkg/lib/util/input.go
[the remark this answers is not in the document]
amisevsk: I can't imagine one, and not trimming likely leads to weirder behaviour than otherwise. This code was originally used for handling login (username and password) where we definitely want to trim white-space, since the read input includes the newline delimiter.

pkg/cmd/init/cmd.go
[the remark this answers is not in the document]
amisevsk: Fixed

pkg/lib/kitfile/generate.go
```diff
@@ -0,0 +1,274 @@
...
+	// Other suffixes
+	".gguf", ".ggml", ".ggmf", ".llamafile", ".onnx",
+}
+
+var docsSuffixes = []string{
+	".md", ".adoc", ".html", ".pdf",
+}
+
+var metadataSuffixes = []string{
+	".json", ".yaml", ".xml", ".csv", ".txt",
```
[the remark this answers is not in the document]
amisevsk: Fixed.

pkg/lib/kitfile/generate.go
amisevsk: Added.

pkg/lib/kitfile/generate.go
amisevsk: Fixed.

pkg/lib/kitfile/generate.go
```diff
@@ -0,0 +1,377 @@
...
+				Path:        filename,
+				Description: "License file",
+			})
+			licenseType, err := detectLicense(filepath.Join(baseDir, filename))
+			if err != nil {
+				output.Debugf("Error determining license type: %s", err)
+				output.Logf(output.LogLevelWarn, "Unable to determine license type")
+			}
+			output.Logf(output.LogLevelTrace, "Detected license %s for license file", detectedLicenseType)
+			detectedLicenseType = licenseType
```
gorkem, line 132: This line should be before the logging.

gorkem (commented):
When I try this on 

```
/scikitlearn-tabular
├── Kitfile
├── data
 │   └── online_boutique_v1.csv
 ── models
 │   └── model.onnx
└── notebooks
    └── ML_MLFlow.ipynb
```

I am getting the following. I was expecting the model.onnx to be added at least.  

```
manifestVersion: 1.0.0
package:
  name: test
  description: this
code:
  - path: .
```

amisevsk:
@gorkem Hm, are you sure you've fetched the latest changes? I rebased on main as part of it so you would need to `git reset --hard ...` to see current state (or delete the branch locally and use `gh pr checkout` again).

I tested with 
```
.
├── data
│   └── online_boutique_v1.csv
├── models
│   └── model.onnx
└── notebooks
    └── ML_MLFlow.ipynb
```
(files are just basic text files containing "test") and I see the Kitfile
```yaml
manifestVersion: 1.0.0
model:
  name: model
  path: models/model.onnx
code:
  - path: notebooks
datasets:
  - path: data
```

pkg/lib/kitfile/generate.go
amisevsk: Woops, good catch! Fixed.

amisevsk:
Figured it out -- I was mixing up dirEntry.Name() and the full path to the directory. If you ran the init command from the directory you want to generate a Kitfile in it would work because it's looking for directories in the current working dir. This fails if kit is run from a different directory (I had tried to test this using `kit init $(pwd)` but that obviously misses this bug).

gorkem (approved):
Needs tests

PR #679, opened 2025-01-08
Add back CookieYes

javisperez:
The GTM related changes should be in and working 🤞 .

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -39,9 +39,9 @@ if (inProd) {
   )
 
   // CookieYes
-  // head.push(
-  //   ['script', { id: 'cookieyes', src: 'https://cdn-cookieyes.com/client_data/43ed933a5c31287e10124caf/script.js' }]
-  // )
+  head.push(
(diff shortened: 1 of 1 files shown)


docs/.vitepress/config.mts
```diff
@@ -39,9 +39,9 @@ if (inProd) {
   )
 
   // CookieYes
-  // head.push(
-  //   ['script', { id: 'cookieyes', src: 'https://cdn-cookieyes.com/client_data/43ed933a5c31287e10124caf/script.js' }]
-  // )
+  head.push(
+    ['script', { id: 'cookieyes', type: 'text/javascript', src: 'https://cdn-cookieyes.com/client_data/43ed933a5c31287e10124caf/script.js' }]
```
gorkem, line 43: Should this use an `async` attribute?

gorkem (approved):

amisevsk (approved):

docs/.vitepress/config.mts
javisperez: not really... their site says to load it sync, i assume that it must block the page so it can prevent the cookies from loading while this loads.

PR #684, opened 2025-01-10
Add huggingface-to-ModelKit flow

amisevsk:
### Description
Add the ability to automatically convert huggingface (and general Git) repositories into ModelKits via the `kit import` command. Provided a repository, `kit import` will

1. Clone (using git-lfs) that repository to a temp directory
2. Generate a Kitfile for it
3. Prompt the user to edit the Kitfile before packing, for an interactive session
4. Package the files to a ModelKit

Since I've encountered issues with the go-git package in the past, this functionality requires git and git-lfs to be installed locally.

### Linked issues
Closes https://github.com/jozu-ai/kitops/issues/564


pkg/lib/git/cleanup.go
```diff
@@ -0,0 +1,45 @@
...
+	"os"
+	"path/filepath"
+)
+
+// CleanGitMetadata removes all git-related files from a repository
+func CleanGitMetadata(dir string) error {
+	filesToClean := []string{
+		".git",
+		".gitattributes",
+		".gitignore",
```
gorkem, line 32: There can be ".gitignore" file can be on the subdirectories. Should those be handled?

amisevsk: This is a good point. Should we scan entire directories in this way? It could become complicated / time-consuming, since we'd have to walk the whole tree just for this, but maybe it's not a huge issue.

I'll add it to the PR.

gorkem (changes requested):
We need to find a way to pass the HF token for repositories that require it. At the moment git goes to interactive mode which is not going to work on CI systems.

pkg/cmd/kitimport/cmd.go
```diff
@@ -0,0 +1,99 @@
...
+
+	cmd.Flags().StringVarP(&opts.tag, "tag", "t", "", "Tag for the ModelKit (default is '[repository]:latest')")
+	cmd.Flags().SortFlags = false
+	return cmd
+}
+
+func runCommand(opts *importOptions) func(*cobra.Command, []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		if err := opts.complete(cmd.Context(), args); err != nil {
+			return output.Fatalf("Invalid arguments: %s", err)
```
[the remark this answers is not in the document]
amisevsk: This one is logged directly, so the uppercase is appropriate

amisevsk:
I'm still working on implementing the download via direct calls to the huggingface API, but for now I've added support for passing tokens to git. This is a little wonky, because git doesn't let us pass a token in via environment or static argument, so what we actually do is use [git config environment variables](https://git-scm.com/docs/git-config#ENVIRONMENT) to add keys to the existing config. With this we're defining a username and credentials helper that just prints out the provided token.

I've also tried to disable any interactive prompt, so the import command should fail with an error if the repository requires auth now.

pkg/lib/git/clone.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// Clone without LFS enabled to get metadata about repo first
+	cloneCmd := exec.Command("git", "clone", "--depth", "1", repo, dest)
+	cloneCmd.Env = os.Environ()
+	cloneCmd.Env = append(cloneCmd.Env, "GIT_LFS_SKIP_SMUDGE=1")
+	cloneCmd.Stderr = os.Stderr
+	cloneCmd.Stdout = os.Stdout
+	output.Infof("Cloning repository %s", repo)
+	if err := cloneCmd.Run(); err != nil {
+		return fmt.Errorf("Error cloning repository: %w", err)
```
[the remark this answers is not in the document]
amisevsk: Fixed

pkg/lib/git/cleanup.go
amisevsk: We're now cleaning .gitattributes and .gitignore in any subpath

gorkem (approved):

PR #685, opened 2025-01-10
Allow specifying a separate target reference to push in `kit push`

amisevsk:
### Description
Allow specifying a separate push target, so that you don't _have_ to tag ModelKits locally before pushing (similar to what podman--but not docker--does).

* Push locally-tagged image to remote (same as what we currently do): 
  ```
  kit push registry.example.com/myorg/mymodel:mytag
  ```
* Push a local image to the same remote without tagging first
  ```
  kit push localmodel:latest registry.example.com/myorg/mymodel:mytag
  ```

### Linked issues
I didn't create an issue for this one yet, but I figured it'd be useful while working on https://github.com/jozu-ai/kitops/pull/684, which by default tags according to the HF repo

### DIFF
--- a/pkg/cmd/push/cmd.go
+++ b/pkg/cmd/push/cmd.go
@@ -47,8 +47,9 @@ kit push registry.example.com/my-model:1.0.0`
 
 type pushOptions struct {
 	options.NetworkOptions
-	configHome string
-	modelRef   *registry.Reference
+	configHome   string
+	srcModelRef  *registry.Reference
+	destModelRef *registry.Reference
 }
 
 func (opts *pushOptions) complete(ctx context.Context, args []string) error {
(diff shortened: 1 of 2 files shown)


pkg/cmd/push/cmd.go
```diff
@@ -33,22 +33,26 @@ import (
 
 const (
 	shortDesc = `Upload a modelkit to a specified registry`
-	longDesc  = `This command pushes modelkits to a remote registry.
+	longDesc  = `This command pushes modelkits from local storage to a remote registry.
 
-The modelkits should be tagged with the target registry and repository before
-they can be pushed`
+If specified without a destination, the ModelKit should be tagged locally before
+pushing.`
```
gorkem, line 39: I think it "must" be tagged otherwise push will fail.

pkg/cmd/push/cmd.go
```diff
@@ -58,17 +62,30 @@ func (opts *pushOptions) complete(ctx context.Context, args []string) error {
...
+	srcRef, extraTags, err := util.ParseReference(args[0])
 	if err != nil {
-		return fmt.Errorf("failed to parse reference: %w", err)
-	}
-	if modelRef.Registry == "localhost" {
-		return fmt.Errorf("registry is required when pushing")
+		return fmt.Errorf("failed to parse reference %s: %w", args[0], err)
 	}
 	if len(extraTags) > 0 {
 		return fmt.Errorf("reference cannot include multiple tags")
```
gorkem, line 70: now that there are 2 references, we should qualify this one as source reference similiar to the dest one.

pkg/cmd/push/cmd.go
```diff
@@ -58,17 +62,30 @@ func (opts *pushOptions) complete(ctx context.Context, args []string) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to parse target reference %s: %w", args[1], err)
+		}
+		if len(extraTags) > 0 {
+			return fmt.Errorf("target reference cannot include multiple tags")
+		}
+		opts.destModelRef = destRef
+	}
+
+	if opts.destModelRef.Registry == "localhost" {
```
gorkem, line 86: `localhost` is a valid use case in local testing. Is this making it impossible?

gorkem (changes requested):

pkg/cmd/push/cmd.go
amisevsk: Updated, but I think the wording here is generally strange -- if you type `kit push quay.io/myorg/mymodel:latest`, what does that mean if you don't have a modelkit tagged `quay.io/myorg/mymodel:latest`?

pkg/cmd/push/cmd.go
amisevsk: I didn't want to complicate the wording too much (so we do distinguish them by labelling the target `target`). If specified with one argument, calling it the source reference doesn't make a lot of sense to me, as it's both the source and destination.

pkg/cmd/push/cmd.go
amisevsk: Nope :)

This check has always been [here](https://github.com/jozu-ai/kitops/blob/60e30a5b2423d3e16c4c99edf527438d6ca79109/pkg/cmd/push/cmd.go#L65-L67) -- it's because we adopted the podman convention of using `localhost` as the registry placeholder in references (so that they're still valid). Local usage isn't impacted, since you would likely tag your ModelKit `localhost:5050`, which doesn't match.

For context, we need a way to store and parse references to ModelKits that don't have full information, while the OCI spec requires both a `registry` and `repository`. For example, completely untagged ModelKits are represented `localhost/_@<digest>`.

gorkem (approved):

PR #688, opened 2025-01-10
Fix race condition when pulling manifest with duplicate digests

amisevsk:
### Description
When a manifests contains two layers with the same digest, pulling that ModelKit can fail if we try to download both (identical) layers simultaneously. The first download to win the race moves the file from the ingest/ directory to blobs/sha256/, and the second fails as the source file does not exist.

To avoid this, we pull each digest only once.

### Linked issues
Closes https://github.com/jozu-ai/kitops/issues/687

### DIFF
--- a/pkg/lib/repo/local/pull.go
+++ b/pkg/lib/repo/local/pull.go
@@ -70,8 +70,16 @@ func (l *localRepo) PullModel(ctx context.Context, src oras.ReadOnlyTarget, ref 
 		return fmt.Errorf("failed to get %s layer: %w", constants.FormatMediaTypeForUser(desc.MediaType), err)
 	}
 	var semErr error
+	// In some cases, manifests can contain duplicate digests. If we try to concurrently pull the same digest
+	// twice, a race condition will cause the pull the fail.
+	pulledDigests := map[string]bool{}
 	for _, pullDesc := range toPull {
 		pullDesc := pullDesc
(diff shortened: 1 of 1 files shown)


pkg/lib/repo/local/pull.go
```diff
@@ -70,8 +70,16 @@ func (l *localRepo) PullModel(ctx context.Context, src oras.ReadOnlyTarget, ref
 		return fmt.Errorf("failed to get %s layer: %w", constants.FormatMediaTypeForUser(desc.MediaType), err)
 	}
 	var semErr error
+	// In some cases, manifests can contain duplicate digests. If we try to concurrently pull the same digest
+	// twice, a race condition will cause the pull the fail.
+	pulledDigests := map[string]bool{}
```
gorkem, line 75: this can be a preallocated map 
```Go
	pulledDigests := make(map[string]bool, len(toPull))
```

gorkem (approved):

pkg/lib/repo/local/pull.go
amisevsk: It could, but in the case we're trying to solve this would be overallocating -- this is a set used to check whether we've seen each element before. With v0.4.0, we can create ModelKits that have the same layer dozens of times (e.g. empty `readme.md` files in different directories)

Issue #692, opened 2025-01-14
missing reference. Check you have write permissions to the organization and repository you are pushing to.


Techtacles:
Using the new dagger release v0.1.0, anytime I try pushing a modelkit to Jozu hub, I get the error 
"missing reference. Check you have write permissions to the organization and repository you are pushing to.".
This was working in the previous version of dagger module. I have a Github Action pipeline that was working previously but now, it's broken. The link to the GitHub Action created previously is https://github.com/Techtacles/kitops-dagger/blob/master/.github/workflows/master.yml

Can you advise the issue?

I have attached a screenshot to this. 
<img width="1009" alt="Screen Shot 2025-01-14 at 6 36 47 PM" src="https://github.com/user-attachments/assets/e685b094-d6ed-40f1-8f5d-873a50082d35" />

gorkem:
I think there is an issue with the reference. A reference must have tags (or SHA) in this case it is just repo. Can you add tags to the action. When it is packed it should have tags something like `jozu.ml/emmanueloffisongetim/llm_repo:latest` and when pushing it should push the `jozu.ml/emmanueloffisongetim/llm_repo:latest`

Techtacles:
I did this. I added the latest tag. It threw the same error. 
I added a screenshot regarding this.
<img width="968" alt="Screen Shot 2025-01-14 at 10 39 27 PM" src="https://github.com/user-attachments/assets/59ea7030-bf24-4663-966d-56ca67031bab" />

Techtacles:
The Modelkit packing works, according to the logs. The issue arises when its time for pushing the modelkit to the registry. Then it throws the write permissions error.

gorkem:
yes, I have checked the logs but without a tag the push command can not find the local reference to the ModelKit you have just packed. Now that you have added the tags we suspect that the dagger module is not configuring the CONFIG_HOME correctly and loosing the local repo between pack and push commands. 

We have also checked the permissions; this is not a permission error and #694 is cleaning up the misleading message.

Techtacles:
I thought about that too but it was working previously with previous dagger modules. Can you check here for more detailed information about the logs? 
https://github.com/Techtacles/kitops-dagger/actions/runs/12795134334/job/35671812023

PR #694, opened 2025-01-15
Clean up error messages on push failures

amisevsk:
### Description
We shouldn't really log a tip about "check that you have write permissions to the organization..." when a push fails unless the push failed for that reason. Our current way of logging is confusing, suggesting there's an issue with the remote even if the push is e.g. missing a tag (or refers to a non-existent tag).

If the returned error is a response error from the registry, log information about the response code (e.g. status code). Otherwise, just log the error to point at what went wrong.

### Linked issues
Related (issue highlights the problem): #692

### DIFF
--- a/pkg/cmd/push/cmd.go
+++ b/pkg/cmd/push/cmd.go
@@ -18,7 +18,9 @@ package push
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
 
 	"kitops/pkg/cmd/options"
 	"kitops/pkg/lib/constants"
@@ -29,6 +31,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"oras.land/oras-go/v2/registry"
+	"oras.land/oras-go/v2/registry/remote/errcode"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/cmd/push/cmd.go
```diff
@@ -138,8 +141,17 @@ func runCommand(opts *pushOptions) func(*cobra.Command, []string) error {
...
 		}
 		desc, err := PushModel(cmd.Context(), localRepo, remoteRepo, opts)
-		if err != nil {
-			return output.Fatalf("Failed to push: %s. Check you have write permissions to the organization and repository you are pushing to.", err)
+		respErr := &errcode.ErrorResponse{}
+		if ok := errors.As(err, &respErr); ok {
+			output.Debugf("Got error pushing: %s", err)
+			errMsg := fmt.Sprintf("Failed to push: got response %d (%s) from remote", respErr.StatusCode, http.StatusText(respErr.StatusCode))
+			switch respErr.StatusCode {
+			case http.StatusUnauthorized:
```
gorkem, line 149: Should we also handle 403?

pkg/cmd/push/cmd.go
```diff
@@ -138,8 +141,17 @@ func runCommand(opts *pushOptions) func(*cobra.Command, []string) error {
 			output.Infof("Pushing %s", opts.srcModelRef.String())
 		}
 		desc, err := PushModel(cmd.Context(), localRepo, remoteRepo, opts)
-		if err != nil {
-			return output.Fatalf("Failed to push: %s. Check you have write permissions to the organization and repository you are pushing to.", err)
+		respErr := &errcode.ErrorResponse{}
+		if ok := errors.As(err, &respErr); ok {
```
gorkem, line 145: respErr is already a pointer so is this passing `**errcode.ErrorResponse`?

gorkem (changes requested):

pkg/cmd/push/cmd.go
amisevsk: Yeah, this is intended -- since errors.As() sets the value of respErr, it needs a reference to the error itself, which leads to two levels of indirection:

* `errcode.ErrorResponse` is the base struct that stores the data
* `*errcode.ErrorResponse` is a reference to the struct, needed to satisfy the `error` interface (which uses pointer receivers)
* `**errcode.ErrorResponse` is needed to allow errors.As() to set the value of the error; this is required since errors.As() wants a pointer to a type that implements the error interface (in this case, `*errcode.ErrorResponse`).

pkg/cmd/push/cmd.go
amisevsk: I considered handling all the common 4xx codes here, but I can't really think of useful additional context to provide apart from e.g.

> `Failed to push: got response 403 (Forbidden) from remote`

I mainly added special handling for 401s to avoid undoing the fix to issue https://github.com/jozu-ai/kitops/issues/510

We can try to fill the normal response codes from the [OCI spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints) but I'm not sure most of them are used except the usual 401 to require an auth challenge.

I believe (but haven't tested) that the oras library will catch and handle responses with the usual OCI codes (e.g. `BLOB_UPLOAD_INVALID`) .

I think @brett-hodges was running into issues with 429 (too many requests), so maybe it makes sense to handle that one specifically?

gorkem (approved):

PR #697, opened 2025-01-17
Review urls to had `trailingSlash` for CF

### LINKED ISSUE #696: Add trailing slash to urls
We need to default all of our urls to he a trailing slash to try and remove CF's 308's.

This includes:

* all docs url's
* in-site links references
* urls in the sitemap.xml
* canonical urls

javisperez:
Fixes #696 by forcing trailing slashes in all our routes.

Related: https://github.com/jozu-ai/pykitops/pull/19


docs/src/docs/pykitops.md
javisperez, line 1: will make the proper pr in PyKitops as well. This is a conflicting file.

javisperez: oh, nevermind, looks like these files are on the KitOps side only, not in PyKitops.

amisevsk (approved):

PR #702, opened 2025-01-20
Add 'diff' command to compare ModelKits

gorkem:
Enable comparison of local and remote ModelKits, displaying shared and unique layers

### DIFF
--- a/pkg/cmd/diff/cmd.go
+++ b/pkg/cmd/diff/cmd.go
@@ -1,0 +1,171 @@
+// Copyright 2024 The KitOps Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package diff
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/spf13/cobra"
+	"oras.land/oras-go/v2/registry"
+
+	"kitops/pkg/cmd/options"
+	"kitops/pkg/lib/constants"
+	"kitops/pkg/output"
+)
+
+const (
+	// Constants for identifying reference types
+	remotePrefix = "remote://"
+	localPrefix  = "local://"
+	//Constants for formatting output tables.
+	layerTableHeadings = "Type    | Digest             | Size"
+	layerTableFormat   = "%-7s | %-18s | %s\n"
+)
+
+type diffOptions struct {
+	options.NetworkOptions
+	configHome string
+	refA       *registry.Reference
+	refB       *registry.Reference
+}
+
+func DiffCommand() *cobra.Command {
+	opts := &diffOptions{}
+	cmd := &cobra.Command{
+		Use:   "diff",
+		Short: "Compare two ModelKits",
+		Long: `Compare two ModelKits to see the differences in their layers.
+		
+		ModelKits can be specified from either a local or from a remote registry.
+		To specify a local ModelKit, prefix the reference with 'local://', e.g. 'local://jozu.ml/foo/bar'.
+		To specify a remote ModelKit, prefix the reference with 'remote://', e.g. 'remote://jozu.ml/foo/bar'.
+		If no prefix is specified, the local registry will be checked first.
+		`,
+		Example: "kit diff modelkit1 modelkit2",
+		RunE:    runCommand(opts),
+	}
+	return cmd
+}
+
+func runCommand(opts *diffOptions) func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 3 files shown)


pkg/cmd/diff/diff.go
```diff
@@ -0,0 +1,115 @@
...
+			delete(layerMapA, layer.Digest.String())
+		} else {
+			result.UniqueLayersB = append(result.UniqueLayersB, layer)
+		}
+	}
+
+	result.UniqueLayersA = make([]ocispec.Descriptor, 0, len(layerMapA))
+	for _, layer := range layerMapA {
+		result.UniqueLayersA = append(result.UniqueLayersA, layer)
+	}
```
amisevsk, line 67: We might want to impose some order on these layers; they'll be iterated randomly from the map by Go.

amisevsk (commented):
A few minor comments

amisevsk (approved):

pkg/cmd/diff/diff.go
```diff
@@ -0,0 +1,127 @@
...
+
+// Helper struct diffResult contains the comparison results between two ModelKits.
+type diffResult struct {
+	SharedLayers  []ocispec.Descriptor
+	UniqueLayersA []ocispec.Descriptor
+	UniqueLayersB []ocispec.Descriptor
+}
+
+// compareManifests compares two OCI manifests and returns the shared and unique layers.
+func compareManifests(manifestA *ocispec.Manifest, manifestB *ocispec.Manifest) *diffResult {
```
amisevsk, line 48: There are two pieces that we're not diffing that could lead to confusing results:

* Any annotations on the manifest (which will happen if e.g. the modelkit is packed by a different version of kit)
* The config for the ModelKit (which could point at e.g. different parameters on the same model data)

pkg/cmd/diff/cmd.go
```diff
@@ -0,0 +1,192 @@
...
+		Use:   "diff",
+		Short: "Compare two ModelKits",
+		Args:  cobra.ExactArgs(2),
+		Long: `Compare two ModelKits to see the differences in their layers.
+		
+		ModelKits can be specified from either a local or from a remote registry.
+		To specify a local ModelKit, prefix the reference with 'local://', e.g. 'local://jozu.ml/foo/bar'.
+		To specify a remote ModelKit, prefix the reference with 'remote://', e.g. 'remote://jozu.ml/foo/bar'.
+		If no prefix is specified, the local registry will be checked first.
+		`,
```
amisevsk, line 62: Annoyingly, using backticks preserves strings _verbatim_, including the leading whitespace :) 

```
kit diff --help
Compare two ModelKits

Description:
  Compare two ModelKits to see the differences in their layers.
  		
  		ModelKits can be specified from either a local or from a remote registry.
  		To specify a local ModelKit, prefix the reference with 'local://', e.g. 'local://jozu.ml/foo/bar'.
  		To specify a remote ModelKit, prefix the reference with 'remote://', e.g. 'remote://jozu.ml/foo/bar'.
  		If no prefix is specified, the local registry will be checked first.
  		

Usage:

```

pkg/cmd/diff/cmd.go
```diff
@@ -0,0 +1,192 @@
...
+
+		output.Infoln("Configurations:")
+		output.Infoln("---------------------------------------")
+		if result.SameConfig {
+			output.Infof("  Configs are identical (Digest: %s)\n\n", diffA.Manifest.Config.Digest[:17])
+
+		} else {
+			fmt.Printf("Configs differ:\n")
+			fmt.Printf("  ModelKit1 Config Digest: %s\n", diffA.Manifest.Config.Digest[:17])
+			fmt.Printf("  ModelKit2 Config Digest: %s\n\n", diffB.Manifest.Config.Digest[:17])
```
amisevsk, line 121: I feel like it would be relatively easy to diff the configs themselves, but that can be a future task.

One note: since we're storing digests in the config for now, any difference in layers will also show up as a difference in configs, even if it's not intended (i.e. we're probably looking to compare Kitfiles here).

pkg/cmd/diff/cmd.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		output.Infoln("Annotations:")
+		output.Infoln("---------------------------------------")
+		if result.AnnotationsMatch {
+			output.Infof("  Annotations are identical \n\n")
+		} else {
+			fmt.Printf("  Annotations does not match\n\n")
```
amisevsk, line 129:
```suggestion
			output.Infof("  Annotations does not match\n\n")
```

pkg/cmd/diff/cmd.go
```diff
@@ -0,0 +1,171 @@
...
+
+	configHome, ok := ctx.Value(constants.ConfigKey{}).(string)
+	if !ok {
+		return fmt.Errorf("default config path not set on command context")
+	}
+	opts.configHome = configHome
+
+	if len(args) < 2 {
+		return fmt.Errorf("this command requires two references. usage: 'kit diff ref1 ref2'")
+	}
```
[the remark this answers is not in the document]
amisevsk: With current changes, this check is no longer necessary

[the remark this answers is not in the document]
gorkem: yeah, forgot about it.

pkg/cmd/diff/cmd.go
gorkem: It is not just that Use and examples are incorrect too. I completely forgot to update. Fixed now

pkg/cmd/diff/cmd.go
gorkem: fixed

amisevsk (approved):
LGTM :+1:

PR #721, opened 2025-01-23
Use 'nano' as the default editor in import if $EDITOR is unset

### LINKED ISSUE #713: Try default editors in `kit import` if `$EDITOR` is not defined
Currently, kit will look to the `EDITOR` environment variable to determine what editor should be used to update the Kitfile before packing. However, not all systems have this environment variable set by default.

We should try a default editor (e.g. `nano` on Linux/MacOS) if the environment variable is unset. If the default editor is not installed, we should print the current message

**Additional context**
Originally reported in https://github.com/jozu-ai/kitops/issues/708

amisevsk:
### Description
On linux and macos, use 'nano' as a default editor, in case the $EDITOR env var is unset. On windows, show the current message, as the default editor there is less clear.

### Linked issues
Closes #713

### DIFF
--- a/pkg/cmd/kitimport/kitimport.go
+++ b/pkg/cmd/kitimport/kitimport.go
@@ -218,7 +218,11 @@ func getEditorName() (string, error) {
 			return path, nil
 		}
 	}
-	// TODO: can we do anything default for Windows platforms?
+	if runtime.GOOS == "windows" {
+		// On current Windows, using notepad will block until the editor (or editor tab)
+		// is closed.
+		return "notepad", nil
+	}
 
 	return "", fmt.Errorf("no editor found")
 }
--- a/pkg/cmd/kitimport/kitimport.go
+++ b/pkg/cmd/kitimport/kitimport.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"kitops/pkg/artifact"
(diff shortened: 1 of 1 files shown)


pkg/cmd/kitimport/kitimport.go
```diff
@@ -204,3 +205,25 @@ func promptToEditKitfile(contextDir, tag string, currentKitfile *artifact.KitFil
...
+
+func getEditorName() (string, error) {
+	// Default: use the editor in the standard $EDITOR environment variable
+	editor := os.Getenv("EDITOR")
+	if editor != "" {
+		return editor, nil
+	}
+	if runtime.GOOS == "linux" || runtime.GOOS == "darwin" {
+		// Check for nano, which should be the default editor on Linux and MacOS
+		checkCommand := exec.Command("command", "-v", "nano")
```
gorkem, line 217: Is this essentially doing `exec.LookPath("nano")`?

pkg/cmd/kitimport/kitimport.go
```diff
@@ -204,3 +205,25 @@ func promptToEditKitfile(contextDir, tag string, currentKitfile *artifact.KitFil
 	}
 	return readExistingKitfile(kitfilePath)
 }
+
+func getEditorName() (string, error) {
+	// Default: use the editor in the standard $EDITOR environment variable
+	editor := os.Getenv("EDITOR")
+	if editor != "" {
+		return editor, nil
```
gorkem, line 213: Would doing an "exec.LookPath()" here to check if the editor really exists makes sense?

gorkem (approved):

pkg/cmd/kitimport/kitimport.go
amisevsk: I don't think so -- if we couldn't find the editor in the path, we'd probably just return an error message saying that, which is what already happens:

```
Would you like to edit Kitfile before packing? (y/N): y
[ERROR] Error running external editor: exec: "nope": executable file not found in $PATH
```

pkg/cmd/kitimport/kitimport.go
amisevsk: Right, forgot that existed :) 

Fixed

PR #740, opened 2025-02-03
Fixes 404's

javisperez:
Because of the `trailing slash` change in url's, some relative .md references are being treated as "folder/file" instead of just "file" (eg. `/docs/get-started/` would append anything to `/docs/get-started/foo` because of the trailing slash).

### DIFF
--- a/docs/src/docs/deploy.md
+++ b/docs/src/docs/deploy.md
@@ -17,7 +17,7 @@ The container is configurable via environment variables:
 
 `$MODELKIT_REF`: The ModelKit to pull (required).
 `$UNPACK_PATH`: Where to unpack the ModelKit (normally you’d want a `volumeMount` here). This is required and will default to `/home/user/modelkit`.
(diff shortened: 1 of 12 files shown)


docs/src/docs/deploy.md
```diff
@@ -17,7 +17,7 @@ The container is configurable via environment variables:
 
 `$MODELKIT_REF`: The ModelKit to pull (required).
 `$UNPACK_PATH`: Where to unpack the ModelKit (normally you’d want a `volumeMount` here). This is required and will default to `/home/user/modelkit`.
-`$UNPACK_FILTER`: Optional filter to limit what is unpacked (e.g., just the model, or model + code). The filter format is the same as the [--filter command line argument](./cli/cli-reference.md) for the Kit CLI.
+`$UNPACK_FILTER`: Optional filter to limit what is unpacked (e.g., just the model, or model + code). The filter format is the same as the [--filter command line argument](../cli/cli-reference.md) for the Kit CLI.
```
bmicklea, line 20: I don't totally understand these slash changes but this is the only line where a `.md` is still used. Should it be a `/`?

javisperez: oh yeah, thank you, this one is a miss.

gorkem (approved):

bmicklea (approved):

PR #780, opened 2025-02-25
Use monolithic uploads when pushing to Google Artifact Registry

### LINKED ISSUE #777: Kit cannot upload large files to Google Artifact Registry
**Describe the bug**
Google artifact registry does not support chunked blob uploads

google-doc-ref

This means that Kit uploads fail if the ModelKit has a layer greater than 100Mb in size. 

Internally, this means we'll likely need to parse push destinations and choose the best way to upload ModelKits per registry.

**To Reproduce**
Generate a ModelKit with a single layer over 100Mb in size and attempt to push it to Google Artifact Registry

**Version**
latest

**Additional context**
Brought up in the discord channel

amisevsk:
### Description
Google Artifact Registry does not support chunked uploads and instead requires monolithic uploads. Handle pushes to the registry similar to how we handle ghcr.io (matching .pkg.dev)

**Note:** I haven't been able to test this directly, yet, as I don't have access to a Google Artifact Registry at the moment. I'll test to double check before merging

### Linked issues
Closes #777

### DIFF
--- a/pkg/lib/repo/remote/upload-format.go
+++ b/pkg/lib/repo/remote/upload-format.go
@@ -16,7 +16,10 @@
 
 package remote
 
-import "kitops/pkg/output"
+import (
+	"kitops/pkg/output"
+	"regexp"
+)
 
 type uploadFormat int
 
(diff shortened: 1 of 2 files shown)


gorkem:
Do we have tests covering upload-format? The list of special treatments may be growing.

amisevsk:
> Do we have tests covering upload-format? The list of special treatments may be growing

Good idea, I'll add some

amisevsk:
Tested on Google Artifact Registry and was able to push a 1GB layer without issue.

I logged in using an access_token, as Kit currently does not support auth helpers.

PR #791, opened 2025-03-04
Explicitly allow OAI bots

javisperez:
Explicitly add `allow` rules for all the OpenAI bots in our robots.txt, even if we're not blocking any bot 😬


gorkem (approved):
LGTM

PR #805, opened 2025-03-22
Change module path in go.mod and update internal imports

gorkem:
This allows kitops to be imported easier. 

- Updated the module declaration in go.mod from "kitops" to "github.com/kitops-ml/kitops".
- Replaced all internal import paths from "kitops/..." to "github.com/kitops-ml/kitops/..." to ensure consistency with the new module path.


amisevsk (approved):
Makes sense, we should've done this a long time ago :)

Issue #786, opened 2025-03-02
KITOPS_CLIENT_CERT and KITOPS_CLIENT_KEY doesn't seem to have an effect


notniknot:
We have a proxy in front of our Artifactory that requires client authentication. Therefore, I set:
- KITOPS_CLIENT_CERT=/.../client.pem
- KITOPS_CLIENT_KEY=/.../key.pem

It results in `[ERROR] Failed to validate the credentials for XXX: GET "https://XXX/v2/": response status code 400: Bad Request`

When I set the `--client` and `--key` flags (pointing to the same files as above) it works as expected.

This suggests, that the env vars are not being respected.

**Version**
1.2.2

bmicklea:
Thanks for flagging this @notniknot, we've pulled this into our current sprint so assuming it isn't too complex we should have it fixed in the next week or two.

amisevsk:
Hi @notniknot -- sorry for the delay!

I'm taking a look now after and I'm unable to reproduce the issue locally. I've tried both with specifying the environment variables with absolute paths and relative paths (to the current working directory) and it seems to work. 

Just to double-check: are you exporting the environment variables in your shell before calling Kit? For example, you need something like
```bash
export KITOPS_CLIENT_CERT="/.../client.pem"
export KITOPS_CLIENT_KEY="/.../key.pem"
kit push [source] [destination]

# or, without exporting

KITOPS_CLIENT_CERT="/.../client.pem" KITOPS_CLIENT_KEY="/.../key.pem" kit push [source] [destination]
```

notniknot:
@amisevsk Yes, I tried with and without exporting.

```bash
KITOPS_CLIENT_CERT="/Users/xxx/.secrets/client.pem" KITOPS_CLIENT_KEY="/Users/xxx/.secrets/key.pem" kit login -u xxx xxx.de
```
or
```bash
export KITOPS_CLIENT_CERT="/Users/xxx/.secrets/client.pem"
export KITOPS_CLIENT_KEY="/Users/xxx/.secrets/key.pem"

kit login -u xxx xxx.de
```

Everytime I get `[ERROR] Failed to validate the credentials for xxx.de: GET "https://xxx.de/v2/": response status code 400: Bad Request`. If I include it via --cert and --key it works.

amisevsk:
Hm, interesting. I'll look into it more.

amisevsk:
[...]

PR #806, opened 2025-03-25
Call networkOptions.complete in login command

amisevsk:
### Description
We had missed a call to 'complete' the network options in the login command. This leads to network options around auth being ignored (specifically, the `KITOPS_CLIENT_CERT` and `KITOPS_CLIENT_KEY` env vars, plus the path to stored credentials.

### Linked issues
Closes https://github.com/kitops-ml/kitops/issues/786


gorkem (approved):
This repeated call to `networkOptions.complete` is error prone. Perhaps we should have a `BaseOptions` that all others derive from in the future that takes care of thigns like this.

PR #822, opened 2025-04-09
Use monolithic PUT to upload large layers to Amazon ECR

### LINKED ISSUE #821: Pushing ModelKits to ECR fails
**Describe the bug**
Attempting to push a ModelKit to ECR will fail with 
```
[ERROR] Failed to push: failed to copy to remote: Created.
```
after the first chunk of a multi-part layer upload is completed. It appears the registry is returning HTTP 201 on the patch request for some reason.

**To Reproduce**
Attempt to push a ModelKit with a layer >100MB in size (to trigger a multi-part upload).

**Version**
v1.3.0

**Additional context**
First mentioned in the discord channel

amisevsk:
### Description
Amazon ECR registries return HTTP 201 after the first chunk in a multi-part upload, which is unexpected and causes the upload to fail. To avoid this, we can default to using a monolithic PUT.

### Linked issues
Closes #821

### DIFF
--- a/pkg/lib/repo/remote/upload-format.go
+++ b/pkg/lib/repo/remote/upload-format.go
@@ -17,8 +17,9 @@
 package remote
 
 import (
-	"github.com/kitops-ml/kitops/pkg/output"
 	"regexp"
+
+	"github.com/kitops-ml/kitops/pkg/output"
 )
 
 type uploadFormat int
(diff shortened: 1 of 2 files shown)


pkg/lib/repo/remote/upload-format.go
```diff
@@ -34,8 +35,9 @@ const (
 )
 
 var (
-	googleArtifactRegistryRegexp  = regexp.MustCompile(`.*\.pkg\.dev$`)
-	googleContainerRegistryRegexp = regexp.MustCompile(`.*\.?gcr.io$`)
+	googleArtifactRegistryRegexp         = regexp.MustCompile(`.*\.pkg\.dev$`)
+	googleContainerRegistryRegexp        = regexp.MustCompile(`.*\.?gcr.io$`)
+	amazonElasticContainerRegistryRegexp = regexp.MustCompile(`.*\.?amazonaws.com(.cn)?`)
```
gorkem: Should this regex also end with anchor($) ?

pkg/lib/repo/remote/upload-format_test.go
```diff
@@ -101,3 +101,30 @@ func TestGetUploadFormatGoogleArtifactRegistry(t *testing.T) {
 		assert.Equal(t, uploadMonolithicPut, uploadFormatLarge, "Large layers should use monolithic put")
 	}
 }
+
+func TestGetUploadFormatAmazonECR(t *testing.T) {
+	testRegistries := []string{
```
gorkem, line 106: We should add a few cases with mixed case URLs.

amisevsk: Good hint -- we're not handling them at all right now :slightly_smiling_face:

gorkem (approved):

Issue #817, opened 2025-04-04
ClusterStorageContainer with `kit://` prefix and Kserve not working


dpnoworyta:
**Use case:**
Our idea is to use KitOps to push model files into OCI compatible repository in AWS ECR and use that model for serving in Kserve EKS. Kserve itself does support ECR (as `oci://` prefix in combination with supported model serving runtimes) but only when used as container repo.

**What I have checked**
I found that there is part of KitOps repo when something similar is already covered here: https://github.com/kitops-ml/kitops/tree/main/build/dockerfiles/KServe. 

First of I think that there is a typo in image path: https://github.com/kitops-ml/kitops/blob/98d3e1e1c29ae86a14446f89eec77d27d0fc5c22/build/dockerfiles/KServe/Dockerfile#L1 because it is pointing to image that it is not reachable:

```
ERROR: failed to solve: ghcr.io/kitops-ml/kit:next: failed to resolve source metadata for 
ghcr.io/kitops-ml/kit:next: failed to authorize: failed to fetch anonymous token: 
unexpected status from GET request to https://ghcr.io/token?scope=repository%3Akitops-ml%2Fkit%3Apull&service=ghcr.io: 
403 Forbidden

docker pull ghcr.io/kitops-ml/kit:next
Error response from daemon: Head "https://ghcr.io/v2/kitops-ml/kit/manifests/next": denied
```

Should that point to `ghcr.io/kitops-ml/kitops:next` like your init container?
https://github.com/kitops-ml/kitops/blob/98d3e1e1c29ae86a14446f89eec77d27d0fc5c22/build/dockerfiles/init/Dockerfile#L3

Assuming the answer is yes, I've tried following further your guide in readme by creating `ClusterStorageContainer` and `InferenceService` that uses `kit://` prefix:

```
apiVersion: serving.kserve.io/v1beta1
kind: InferenceService
metadata:
  name: sklearn-movies-kitops-custom-storage
  namespace: kserve
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
      storageUri: "kit://<ACCOUNT_ID>.dkr.ecr.eu-central-1.amazonaws.com/kserve/models:kitpacked-model-1"
```

**Issue**
Upon deploying inference manifest, there are few resources created, one of which is ReplicaSet that below log is from:
```
'Pod "sklearn-movies-kitops-custom-storage-predictor-856bfd8966-hklgn"
[...]

PR #825, opened 2025-04-11
feature: Conditional login to AWS ECR using IRSA based on ENVs

dpnoworyta:
### Description
This PR introduces possibility to set and use AWS IRSA to perform authentication within Kubernetes. 
It is conditional on environment variable specific flags: `AWS_ROLE_ARN` which is automatically set by Kubernetes when service account used for `InferenceService` has `eks.amazonaws.com/role-arn: arn:aws:iam::<ACCOUNT_ID>:role/<ROLE_NAME>` annotation set and `AWS_ECR_REGION` which points to ECR region holding wanted artifacts.

Additionally - because Kitfile unpack might be omitted when user sets `UNPACK_FLAGS` - preview of Kitfile in entrypoint is also conditional.

### Linked issues
fixes https://github.com/kitops-ml/kitops/issues/817

### DIFF
--- a/build/dockerfiles/Dockerfile
+++ b/build/dockerfiles/Dockerfile
@@ -21,7 +21,7 @@ ENV USER_ID=1001 \
     HOME=/home/user/
 
 RUN apk --no-cache upgrade && \
-    apk add --no-cache bash git git-lfs && \
+    apk add --no-cache aws-cli bash git git-lfs && \
     git-lfs install --system && \
     mkdir -p /home/user/ && \
     adduser -D $USER_NAME -h $HOME -u $USER_ID
--- a/build/dockerfiles/KServe/entrypoint.sh
+++ b/build/dockerfiles/KServe/entrypoint.sh
@@ -6,6 +6,8 @@ echo "Binary version info:"
 kit version
 
 read -r -a UNPACK_FLAGS <<< "$KIT_UNPACK_FLAGS"
+read -r AWS_ECR_REGION <<< "$AWS_ECR_REGION"
+read -r AWS_ROLE_ARN <<< "$AWS_ROLE_ARN"
 
 if [ $# != 2 ]; then
   echo "Usage: entrypoint.sh <src-uri> <dest-path>"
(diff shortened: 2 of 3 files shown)


amisevsk:
Thanks for the PR @dpnoworyta! I'll add @gorkem to this PR as he was looking into using the IRSA token on AWS this week

build/dockerfiles/Dockerfile
```diff
@@ -21,7 +21,7 @@ ENV USER_ID=1001 \
     HOME=/home/user/
 
 RUN apk --no-cache upgrade && \
-    apk add --no-cache bash git git-lfs && \
+    apk add --no-cache aws-cli bash git git-lfs && \
```
amisevsk, line 24: Do you know how much the image size is increased due to adding aws-cli?

build/dockerfiles/KServe/entrypoint.sh
```diff
@@ -15,9 +17,18 @@ fi
 REPO_NAME="${1#kit://}"
 OUTPUT_DIR="$2"
 
+if [ -n "$AWS_ROLE_ARN" ]; then
+  AWS_ACCOUNT_ID=$(echo "$AWS_ROLE_ARN" | cut -d: -f5)
+  echo "Logging into AWS ECR $AWS_ACCOUNT_ID.dkr.ecr.$AWS_ECR_REGION.amazonaws.com"
+  aws ecr get-login-password --region $AWS_ECR_REGION | kit login --username AWS --password-stdin $AWS_ACCOUNT_ID.dkr.ecr.$AWS_ECR_REGION.amazonaws.com
+fi
```
amisevsk, line 24: This will only work for the standard AWS endpoints (i.e. no FIPS, or China-based endpoints), but I can't think of a good way to support those anyways. Is there a clean way to document this in the readme?

amisevsk (commented):
Generally looks good to me, but I'm not currently set up to test. I'll let Gorkem have the final say here.

gorkem:
I was working on this and enabling a more generic authentication. For IRSA I was thinking that it would be possible to use IRSA token that is mounted to `/var/run/secrets/eks.amazonaws.com/serviceaccount/token`. PR #827 is taking that approach. Does that work for you?

dpnoworyta:
> I was working on this and enabling a more generic authentication. For IRSA I was thinking that it would be possible to use IRSA token that is mounted to `/var/run/secrets/eks.amazonaws.com/serviceaccount/token`. PR #827 is taking that approach. Does that work for you?

I'm guessing that you haven't tested that, correct? Because looking at this I would say it will not work. In fact I know it won't work. This is because token mounted in `/var/run/secrets/eks.amazonaws.com/serviceaccount/token` is not an authorization token to be used as you proposed, but token that is used to setup trust relationship and based on that AWS STS will grant access to AWS resources (different token).

build/dockerfiles/Dockerfile
dpnoworyta: + ~175 MB

gorkem:
You are right. I have not tested the AWS IRSA bits on #827 yet that is next. Just wanted to put it out to get some ideas.

dpnoworyta:
Got it - without implementation inside core KitOps code, my idea was to leverage `aws-cli` as a low hanging fruit to do the heavy lifting

dpnoworyta:
nah, just didn't wanted to push any unwanted agenda 😉

dpnoworyta:
@gorkem is there a time frame when release process is happening by you guys? I would like inform my supervisors when we could use properly supported feature of KitOps 🙂

gorkem (approved):

gorkem:
@dpnoworyta thank you for the contribution. These changes should be available on `ghcr.io/kitops-ml/kitops-kserve:next` momentarily.

amisevsk:
We can also cut a v1.4.0 release this week to avoid requiring the rolling `next` tag. I don't think we have anything else coming down the pike that's worth holding off a release for.

gorkem:
I want to rebase https://github.com/kitops-ml/kitops/pull/827 on top of this though.

dpnoworyta:
Thank you @gorkem !

PR #827, opened 2025-04-11
Add authentication for KServer ClusterStorageContaier

gorkem:
Updates the entrypoint.sh do that it performs a kit login if KIT_USER and KIT_PASSWORD are set or AWS IRSA is enabled.

### DIFF
--- a/build/dockerfiles/KServe/entrypoint.sh
+++ b/build/dockerfiles/KServe/entrypoint.sh
@@ -17,10 +17,17 @@ fi
 REPO_NAME="${1#kit://}"
 OUTPUT_DIR="$2"
 
+if [ -n "$KIT_USER" ] && [ -n "$KIT_PASSWORD" ]; then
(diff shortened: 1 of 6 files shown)


build/dockerfiles/KServe/entrypoint.sh
```diff
@@ -6,8 +6,6 @@ echo "Binary version info:"
 kit version
 
 read -r -a UNPACK_FLAGS <<< "$KIT_UNPACK_FLAGS"
-read -r AWS_ECR_REGION <<< "$AWS_ECR_REGION"
-read -r AWS_ROLE_ARN <<< "$AWS_ROLE_ARN"
```
amisevsk, line 10: I think these changes were dropped in the rebase -- the AWS IRSA handling is still useful outside the more general `login` use case.

amisevsk: Ah I missed the change below, sorry

amisevsk (approved):

PR #841, opened 2025-04-21
Add kserve integration documentation

gorkem:
Add a new section for KServe integration in the documentation. Moves existing CICD and MLflow documentation to the integrations folder.


docs/src/docs/integrations/kserve.md
```diff
@@ -0,0 +1,96 @@
+---
+description: Learn how to integrate KitOps ModelKits with KServe using a ClusterStorageContainer.
+---
+# Integrating KitOps with KServe
+
+KitOps can be used as a `ClusterStorageContainer` in [KServe](https://kserve.github.io/website/master/) to serve ModelKits directly in InferenceServices. This guide shows how to install, configure, and build the KitOps container for KServe.
```
bmicklea, line 6: I think this line could use some "so what" clarification for folks who don't know what a ClusterStorageContainer gets them. Does this make things easier? Faster? More secure?

bmicklea (changes requested):

docs/src/docs/integrations/kserve.md
gorkem: It is the way KServe retrieves models for inference. It is just how KServe abstracts model storage. So what is "you are using KServe" 😺

gorkem: Added an intro paragraph.

bmicklea (approved):

PR #842, opened 2025-04-21
Fix unpacking referenced ModelKits when filters are used

### LINKED ISSUE #840: Kit unpack with a ModelKit that contains a reference is pulling the entire referenced ModelKit
**Describe the bug**
When unpacking a ModelKit using filters, kit will only download the necessary layers as specified by the filters. This allows e.g. unpacking only the README.md without pulling the entire (potentially very large) model weights.

However, if the to-be-unpacked ModelKit contains a reference for its ModelKit path, the unpack process will pull the entire referenced ModelKit before unpacking.

**To Reproduce**
Compare the behaviour of 
```bash
kit unpack jozu.ml/jozu/fine-tuning:latest --filter=code:README.md
```
versus
```
kit unpack jozu.ml/jozu/llama3-8b:8B-instruct-q4_0 --filter=code:README.md
```

**Version**
v1.5.0

**Additional context**
Originally mentioned in the discord channel.

amisevsk:
### Description
Fix how unpack filters are handled when unpacking referenced ModelKits. Previously, since we only want the model from a referenced ModelKit, we were using a `--filter=model` filter to avoid unpacking other elements of the ModelKit. However, this filter was overwriting any user-provided filters through the `--filter` flag.

This PR also fixes a few minor linter warnings I came across.

### Linked issues
Closes #840

### DIFF
--- a/pkg/cmd/dev/cmd.go
+++ b/pkg/cmd/dev/cmd.go
@@ -17,7 +17,6 @@ package dev
 
 import (
 	"net"
-	"os"
 	"strconv"
 
 	"github.com/kitops-ml/kitops/pkg/lib/harness"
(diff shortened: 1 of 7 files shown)


gorkem (approved):
lgtm

PR #843, opened 2025-04-21
Update all docs dependencies

### LINKED ISSUE #838: Update all docs dependencies
Dependabot has opened PR's for non-patch updates for the DOCS, so instead of just merging, we need to test it and make sure that no breaking change goes in untested. This PR should override the dependabots findings and PR's.

javisperez:
This PR fixes #838 by updating all dependencies for the `/docs` path.

TailwindCSS was upgraded to v4 which means:

- major breaking changes between v3 and v4
- got rid of the postcss plugin and used the vite version of tailwindcss
- moved all the custom values from tailwind.config.js to tailwind.css
- no js api at all, all css variables and removed some `@apply` rules in the css for just the direct values
- updates in elements classes and `!important` flags

There is a problem tho; because Vitepress is not using native css layers, and tailwindcss v4 is, there are some conflicts and overriding happening between the two, so, some classes and preflight definitions needs to be loaded in the same level as the Vitepress theme. Apparently (https://github.com/vuejs/vitepress/issues/4425) this is for old browsers compatibility with Vitepress but it might be "fixed" in the future. In the meantime, i just skipped the `base` layer definition for preflight and dumped the whole base content directly into `tailwind.css`.


amisevsk (approved):

docs/.vitepress/theme/components/Accordion.vue
```diff
@@ -11,7 +11,7 @@ const props = withDefaults(defineProps<{
 </script>
 
 <template>
-<Component :is="props.tag" class="border-b border-opacity-20 border-gold py-10 px-4 p2 text-gray-06">
+<Component :is="props.tag" class="border-b border-gold/20 !py-10 !px-4 p2 !text-gray-06">
```
gorkem, line 14: Is this supposed to be <component>  with lowercase?

gorkem (approved):

docs/.vitepress/theme/components/Accordion.vue
javisperez: hmm not sure what exactly you're talking about (the tag name, the class, or something else) but everything is in the correct case: `Component`, capital C as per common practice with custom element / components (not required, just a practice that most dev follows, including me), and the css classes all lowercase, hyphen separated as per xhtml / html5 standard, not required either, just common.

PR #849, opened 2025-04-22
Fix the build GH workflow for the dev-mode

javisperez:
There's a conflict between the dev-mode pnpm version and the docs pnpm version in the build GH workflow, this PR should fix it by letting both use the same version as the docs' packageManager version which seems to be working fine (8 vs 10). 🤞


.github/workflows/build-deploy-docs.yaml
```diff
@@ -1,70 +1,55 @@
-name: Build and deploy the docs
+name: Build the docs
```
amisevsk, line 1: File name is out of date (build-deploy-docs.yaml)

.github/workflows/build-deploy-docs.yaml
```diff
@@ -1,70 +1,55 @@
...
 
 on:
-  # Runs on pushes targeting the `main` branch, and only for the `docs/**` path.
-  push:
-    branches: [main]
+  # Runs pull requests and only for the `docs/**` path.
+  pull_request:
```
amisevsk, line 5: Do we still want to restrict this to only run on main PRs?

amisevsk (approved):
LGTM, assuming none of the previous build-deploy-docs job artifacts are still needed.

.github/workflows/build-deploy-docs.yaml
javisperez: I renamed it to `build-docs.yaml` but then reverted it because i dont know if there's something else i should do if i renamed this, but will rename it again.

gorkem (approved):

PR #865, opened 2025-05-06
Generate llm.txt at build time

javisperez:
This PR:
* adds a vitepress plugin to generate llm.txt and llm-full.txt dynamically from the content.
* enables the `dead link` warning when building.


bmicklea (approved):
LGTM

PR #873, opened 2025-05-16
Add documentation for initContainer deployment

gorkem:
Adds documentation that describes how to deploy ModelKits using initContainers in Kubernetes.

### DIFF
--- a/docs/src/docs/integrations/k8s-init-container.md
+++ b/docs/src/docs/integrations/k8s-init-container.md
@@ -28,6 +28,8 @@ volumes:
     emptyDir: {}
 ```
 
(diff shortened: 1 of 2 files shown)


docs/src/docs/integrations/k8s-init-container.md
```diff
@@ -0,0 +1,193 @@
...
+
+## 1. Define a Shared Volume
+
+Use an `emptyDir` volume so both containers share the unpacked files:
+
+```yaml
+volumes:
+  - name: modelkit-storage
+    emptyDir: {}
+```
```
amisevsk, line 29: Maybe we need a note/warning about ephemeral storage size, as this could fail kind of weirdly for larger models.

gorkem: Would a link to https://kubernetes.io/docs/concepts/storage/volumes/#emptydir work?

PR #882, opened 2025-05-20
Add `--ignore-existing|-i` flag to kit unpack

### LINKED ISSUE #874: Init container fails on existing PVC contents (no overwrite)
When using the Kit init container backed by a PersistentVolumeClaim, it will unpack a ModelKit into the mounted volume—but if that PVC already contains files from a previous run, the `kit unpack` step exits with an error because it does not overwrite by default.

## Proposed Solutions

- Support KIT_UNPACK_FLAGS
Mimic the KServe container's pattern: introduce an env var like `KIT_UNPACK_FLAGS` so users can pass arbitrary flags (e.g. --overwrite) to the unpack step without rebuilding the image.
 
- Change kit behaviour so that it compares file SHAs and only fails the unpack when the source and destination differ, otherwise skips the copying of the file.

amisevsk:
### Description
Add flag
```
  -i, --ignore-existing      Skip unpacking files if a file with that name already exists
```

to kit unpack. This allows kit unpack to be run idempotently (i.e. only unpacking files that do not currently exist) and is intended for use in e.g. init containers that unpack into persistent volumes.

This commit also changes how existing Kitfiles are processed:

* if the overwrite flag is provided, the existing Kitfile is overwritten as before
* otherwise, we compare the existing Kitfile and do not report and error if the existing Kitfile matches the to-be-unpacked one

Additionally, add support for `$EXTRA_FLAGS` env var in the kit init container, which can be  used to specify arbitrary flags for the unpack command as we do for the KServe container, with the following caveats:

* By default, we use `--ignore-existing` for the init container; this can be disabled by setting some other value for `$EXTRA_FLAGS`
* If `$EXTRA_FLAGS` contains `--filter` or `--dir` flags, we exit with an error

### Linked issues
Closes #874

### DIFF
--- a/build/dockerfiles/init/entrypoint.sh
+++ b/build/dockerfiles/init/entrypoint.sh
@@ -5,6 +5,7 @@ set -e
 # Kit configuration via env vars:
 #   - UNPACK_PATH   : Where to unpack the modelkit (default: /home/user/modelkit)
 #   - UNPACK_FILTER : What to unpack from the modelkit (default: model)
+#   - EXTRA_FLAGS   : Additional flags for the unpack operation (default: --ignore-existing)
 #   - MODELKIT_REF  : The modelkit to unpack -- required!
 UNPACK_PATH=${UNPACK_PATH:-/home/user/modelkit/}
 UNPACK_FILTER=${UNPACK_FILTER:-model}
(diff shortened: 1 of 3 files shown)


amisevsk:
@gorkem if you have time, please take a look; I'm not super happy with how this PR is handling flags for the init container use case.

gorkem (approved):

PR #901, opened 2025-06-16
Make sure directories exist when saving ModelKit to local storage

amisevsk:
### Description
As the 'save ModelKit' flow has optimizations to prevent unnecessary copying (i.e. moving blobs directly into the correct path within local storage), we need to make sure these directories exist before attempting to move the file. Otherwise, packaging ModelKits will use to copy flow for uninitialized local storage.

### Linked issues
I noticed a message when running Kit e.g. in GitHub actions, where the local store is not initialized:

```
[DEBUG] Failed to move temp file into storage (will copy instead): rename /home/runner/work/.kitops-build/cache/pack/kitops_layer_608598814 /home/runner/work/.kitops-build/storage/blobs/sha256/9fa9eafc5ccabf8d84b513c7f0141ad16b29589f249f81116c017ab4a5228ceb: no such file or directory
```

This is because the `$KITOPS_HOME/storage/blobs/sha256` directory does not initially exist.


pkg/lib/repo/local/registry.go
```diff
@@ -120,87 +121,95 @@ func GetAllLocalRepos(storagePath string) ([]LocalRepo, error) {
...
 }
 
-func (l *localRepo) GetTags(desc ocispec.Descriptor) []string {
-	return l.localIndex.listTags(desc)
+func (lr *localRepo) GetTags(desc ocispec.Descriptor) []string {
+	return lr.localIndex.listTags(desc)
+}
+
+func (lr *localRepo) EnsureDirs() error {
+	path := filepath.Join(lr.storagePath, ocispec.ImageBlobsDir, "sha256")
```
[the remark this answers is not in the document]
amisevsk: A fair point, though currently OCI only really does sha256.

pkg/lib/repo/local/registry.go
```diff
@@ -120,87 +121,95 @@ func GetAllLocalRepos(storagePath string) ([]LocalRepo, error) {
 }
 
 // GetRepo returns the registry and repository for the current OCI store.
```
gorkem:
```suggestion
// GetRepoName returns the string representation of <registry>/<repository> for the current local repo.
```

pkg/lib/repo/local/registry.go
```diff
@@ -120,87 +121,95 @@ func GetAllLocalRepos(storagePath string) ([]LocalRepo, error) {
...
-	return l.localIndex.delete(target)
+	return lr.localIndex.delete(target)
 }
 
-func (l *localRepo) Exists(ctx context.Context, target ocispec.Descriptor) (bool, error) {
+func (lr *localRepo) Exists(ctx context.Context, target ocispec.Descriptor) (bool, error) {
 	if target.MediaType == ocispec.MediaTypeImageManifest {
-		return l.localIndex.exists(target), nil
+		return lr.localIndex.exists(target), nil
 	} else {
```
gorkem, line 154: Redundant else block

amisevsk:
I've added trace logging to registry actions; currently we're using the 'safe' variant of our output methods to make sure we don't accidentally interfere with progress bars.

PR #907, opened 2025-06-23
GKE kserve clusterstoragecontainer integration

tmbochenski:
### Description
Native support for auth to GAR on GKE using Workload Identity Federation in KServe ClusterStorageContainer.

### DIFF
--- a/build/dockerfiles/KServe/entrypoint.sh
+++ b/build/dockerfiles/KServe/entrypoint.sh
@@ -23,7 +23,7 @@ elif [ -z "$KIT_USER" ] && [ -z "$KIT_PASSWORD" ] &&  [ -n "$AWS_ROLE_ARN" ]; th
   AWS_ACCOUNT_ID=$(echo "$AWS_ROLE_ARN" | cut -d: -f5)
   echo "Logging into AWS ECR $AWS_ACCOUNT_ID.dkr.ecr.$AWS_ECR_REGION.amazonaws.com"
   aws ecr get-login-password --region $AWS_ECR_REGION | kit login --username AWS --password-stdin $AWS_ACCOUNT_ID.dkr.ecr.$AWS_ECR_REGION.amazonaws.com
-elif [ -z "$KIT_USER" ] && [ -z "$KIT_PASSWORD" ] &&  [ "$GCP_WIF" == "true" ]; then
+elif [ -z "$KIT_USER" ] && [ -z "$KIT_PASSWORD" ] && [ "$GCP_WIF" == "true" ]; then
   if [ -z "$GCP_GAR_LOCATION" ]; then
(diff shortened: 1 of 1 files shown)


build/dockerfiles/KServe/entrypoint.sh
```diff
@@ -23,6 +23,13 @@ elif [ -z "$KIT_USER" ] && [ -z "$KIT_PASSWORD" ] &&  [ -n "$AWS_ROLE_ARN" ]; th
   AWS_ACCOUNT_ID=$(echo "$AWS_ROLE_ARN" | cut -d: -f5)
   echo "Logging into AWS ECR $AWS_ACCOUNT_ID.dkr.ecr.$AWS_ECR_REGION.amazonaws.com"
   aws ecr get-login-password --region $AWS_ECR_REGION | kit login --username AWS --password-stdin $AWS_ACCOUNT_ID.dkr.ecr.$AWS_ECR_REGION.amazonaws.com
+elif [ -z "$KIT_USER" ] && [ -z "$KIT_PASSWORD" ] && [ "$GCP_WIF" == "1" ]; then
```
amisevsk: Minor nit: I prefer this style of bash variable to be just the string "true" instead of "1". I feel it looks better in yaml in e.g. a k8s deploy.

```suggestion
elif [ -z "$KIT_USER" ] && [ -z "$KIT_PASSWORD" ] && [ "$GCP_WIF" == "true" ]; then
```

amisevsk (commented):
Looks good to me, one comment. Thanks @tmbochenski!

build/dockerfiles/KServe/Dockerfile
```diff
@@ -2,8 +2,17 @@ ARG KIT_BASE_IMAGE=ghcr.io/kitops-ml/kitops:next
...
 
 USER 0
+
 RUN apk --no-cache upgrade && \
-    apk add --no-cache aws-cli
+    apk add --no-cache aws-cli curl
+
+RUN mkdir -p /usr/local/gcloud && \
+    curl -s https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz | tar -xz -C /usr/local/gcloud && \
+    /usr/local/gcloud/google-cloud-sdk/install.sh --quiet --usage-reporting false
```
amisevsk: I just noticed this is inflating the image size quite significantly -- 240MB on main to 825MB with this PR. Can we restrict which components are installed to reduce size?

tmbochenski: Unfortunately, it is not possible/easy to reduce the size of gcloud without any “hacky” solution. However, I found an alternative way to get the token that does not increase the image size. I will update the PR.

PR #910, opened 2025-06-24
auth: Docker credential store fallback

### LINKED ISSUE #213: Kit CLI should be able to use same credentials as docker
kit CLI should be able to share the credentials with docker. If a user is authenticated with a registry using docker, they should be able to push to that registry without re-authenticating.

mohammedahmed18:
fixes #213

combine the existing credential store with docker store fallback (for Get method only)


amisevsk (approved):
LGTM, thanks @mohammedahmed18!

PR #915, opened 2025-07-01
:zap: Implement parallel chunking for large file downloads

anxuanzi:
## Description
This PR introduces a significant performance improvement for pulling large files by implementing a parallel chunking download strategy. It also optimizes buffer usage for all downloads to enhance throughput.

The key changes include:

- Parallel Downloads for Large Files: A new `downloadFileInChunks` function has been added. This function splits files larger than a certain threshold (`largeLayerThreshold`, currently 100MB) into chunks and downloads them concurrently. This approach is designed to maximize bandwidth utilization, leading to faster downloads, especially on high-speed internet connections.

- Optimized Buffer Usage: In the standard download functions (`resumeAndDownloadFile` and `downloadFile`), `io.Copy` has been replaced with `io.CopyBuffer`, utilizing a larger buffer (2MB). This change reduces the number of system calls and can improve I/O throughput.

- Dynamic Download Strategy: The pullNode function now intelligently decides which download strategy to use. For large files on servers that support range requests, it will use the new parallel chunking method. Otherwise, it will fall back to the existing resumable or basic download methods.

These enhancements will provide a noticeably faster and more efficient experience for users when pulling large models.

### DIFF
--- a/pkg/lib/repo/local/pull.go
+++ b/pkg/lib/repo/local/pull.go
@@ -20,15 +20,16 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
-	"io/fs"
-	"os"
-	"path/filepath"
-
 	"github.com/kitops-ml/kitops/pkg/cmd/options"
 	"github.com/kitops-ml/kitops/pkg/lib/constants"
 	"github.com/kitops-ml/kitops/pkg/lib/repo/util"
 	"github.com/kitops-ml/kitops/pkg/output"
+	"io"
+	"io/fs"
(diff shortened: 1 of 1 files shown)


anxuanzi:
@amisevsk mind to review this?

amisevsk:
I'll take a look when I have a moment

bmicklea:
I'm guessing this is superseded by PR #926 ?

anxuanzi:
> I'm guessing this is superseded by PR #926 ?

Yep, #926 is better~

PR #925, opened 2025-07-16
Fix wording on Linux install docs

amisevsk:
### Description
The Linux install steps refer to Mac chip type.


bmicklea (approved):
Good catch Angel - thanks!

PR #934, opened 2025-07-31
Better shell completions

### LINKED ISSUE #79: Support shell completions for commands
### Description

Add completions for operations that work on local storage:

* `kit push <modelkit>` -- complete with ModelKits (registry/repository:tag) stored locally
* `kit unpack <modelkit>` -- same as above
* `kit remove <modelkit>` -- same as above
* `kit tag <source> <dest>` -- complete for source ModelKit; maybe for destination as well?
* `kit list` -- complete for local repositories (i.e. where a registry is not included)

mohammedahmed18:
Fixes #79

This update includes two main changes:

1. **Auto-installation for shell completions** — currently supports Zsh, Bash, and Fish.

 https://github.com/user-attachments/assets/e98bdb54-f111-470f-b1fc-b4eda0365d67

2. **Command-specific completions**, as described in the issue.

https://github.com/user-attachments/assets/964a3aa0-6c9e-41a9-b2a5-2d41c5f86c16

> Let me know if you'd like to tweak anything!


cmd/root.go
```diff
@@ -138,19 +136,28 @@ func RunCommand() *cobra.Command {
 	cobra.AddTemplateFunc("indent", indentBlock)
 	cobra.AddTemplateFunc("sectionHead", sectionHead)
 	cobra.AddTemplateFunc("ensureTrailingNewline", ensureTrailingNewline)
+	err := installShellCompletions(cmd, opts)
+	if err != nil {
+		output.Debugf("Failed to install shell completions: %s", err)
```
gorkem, line 141: Errors are only logged at debug level, but shell completion installation failures should be more visible. Silent failures could confuse users who expect completions to work.

cmd/completions.go
```diff
@@ -0,0 +1,117 @@
...
+	content, err := os.ReadFile(pathOfShellRC)
+	if err != nil {
+		return err
+	}
+	sourceCmd := "source " + pathOfCompletionFile
+	if strings.Contains(string(content), sourceCmd) {
+		return nil // already present
+	}
+
+	shellRC, err := os.OpenFile(pathOfShellRC, os.O_APPEND|os.O_WRONLY, 0644)
```
gorkem: There's a race condition between checking if the source command exists and writing to the file. Multiple concurrent executions could result in duplicate entries.

cmd/root.go
```diff
@@ -170,6 +177,12 @@ func Execute() {
 	}
 }
 
+func onConfigHomeError(err error) {
+	output.Fatalf("Failed to read base config directory")
```
gorkem, line 181: output.Fatalf() should terminate execution, but subsequent output.Infof() calls can make it confusing.

cmd/completions.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return err
+	}
+	defer shellRC.Close()
+
+	_, err = shellRC.WriteString("\n# Kitops completions\nsource " + pathOfCompletionFile + "\n")
```
gorkem: This is not safe If an attacker can control the configHome path (through the --config flag or KITOPS_HOME environment variable), they could inject malicious shell commands.  Properly escape or quote the path.

gorkem (changes requested):
Each command initializes completions by reading all local repos, even if completions aren't used. This adds startup latency ( I have alot of local repos). Lazy load completions - only read local repos when tab-completion is actually triggered

cmd/root.go
mohammedahmed18: maybe not doing this because if the user has a problem with the setup, for example (unsupported shell), we don't want to keep showing this message on every command he run, it would be a bad experience for him, what do you think

cmd/root.go
mohammedahmed18: not sure actually that was already there but output.Fatalf() doesn't exit
so I manually `os.exit(1)`

mohammedahmed18:
going to fix file lock for windows later today

update: ignoring shell windows completions for now, maybe will add it in a separate PR

amisevsk:
Thanks @mohammedahmed18 for the contribution. However, I think this approach to completions has a number of issues that make it unsuitable. Have a look at PR #963, which I think is a fair bit cleaner and replaces this one.

* We shouldn't have to adjust how we're passing data around (the configHome parameter), and making some commands require it explicitly while others get it from context is a little messier than I'd like
* I don't think we should be installing completions for users; it's error prone and dependent on too many factors to reliably do automatically
* We should avoid having platform-dependent code wherever possible
* Generating the completions is a little messy -- they're not static completions in the current revision, as they use the `ValidArgsFunction`. Additionally, we don't need to check for prefixes in the completion generator.
* I think it's cleaner to not generate a list of _all_ modelkits and tags when requesting completions -- it's just too much data at once. I took a different approach in #963 if you're interested.

mohammedahmed18:
@amisevsk yeah, fair points
Thanks for th review though

PR #935, opened 2025-08-01
Why not using the latest tag as default ?

mohammedahmed18:


amisevsk (approved):
I think this is a good idea, considering how many users I've seen run into the same issue.

PR #936, opened 2025-08-01
Correct username prompt for logging in to jozu.ml

mohammedahmed18:
jozu-login


gorkem:
May be this would work better if it was a url-to-label map that could be utilized for other registries.

gorkem:
I am closing this because this may eventually cause us to special case every known registry.

PR #944, opened 2025-08-12
feat(dev): add support for ModelKit references in dev command

### LINKED ISSUE #743: Kit dev mode should be usable with a ModelKit reference
Currently, `kit dev start` requires a local directory to run, which means that existing ModelKits must be unpacked before using `kit dev`. While the current implementation is suitable for the intended use case (e.g. testing fine-tuning in development), it is cumbersome for just trying a ModelKit locally.

We should be able to provide a ModelKit reference to `kit dev`. In this case, kit should extract the ModelKit to a temporary directory and serve it from there.

N/A

**Additional context**
N/A

gorkem:
- Add ModelKit reference detection and extraction to dev command
 - Extract ModelKits directly to cache directory
 - Support both directory paths and ModelKit references (e.g., registry.com/model:tag)
 - Include model and kitfile filters for dev-specific extraction
 - Add cleanup for extracted ModelKit files on exit
 - Update command usage and examples to show ModelKit reference support

  Enables `kit dev start <modelkit-reference>` alongside existing directory support,
  allowing users to run dev servers directly from ModelKit references without
  manual unpacking.

fixes #743


pkg/cmd/dev/dev.go
```diff
@@ -17,21 +17,55 @@ package dev
...
+				if cleanupErr := options.cleanup(); cleanupErr != nil {
+					output.Debugf("Failed to cleanup cache directory: %v", cleanupErr)
+				}
+				cleanupDone <- true
+			case <-cleanupDone:
+			case <-signalCtx.Done():
+				return
+			}
+		}()
+	}
```
amisevsk: I'm not sure this will work as expected -- we set up signal handlers and run a goroutine, but the overall process is terminating; once `kit dev start` completes, it will exit, tearing down signal handlers and killing any goroutines (normally, you have to wait on them).

pkg/cmd/dev/dev.go
```diff
@@ -88,6 +124,15 @@ func stopDev(_ context.Context, options *DevBaseOptions) error {
...
 	}
+
+	// Clean up cache directory
+	cacheDir := filepath.Join(options.configHome, "dev-models", "current")
+	if err := os.RemoveAll(cacheDir); err != nil {
+		output.Debugf("Failed to clean up cache directory: %v", err)
+	} else {
+		output.Infof("Cleaned up cache directory")
+	}
+
```
amisevsk: This fails to clean the cache directory if the dev server has crashed / been terminated in some other way -- e.g. if I try to `kit dev start` with an onnx ModelKit, the dev server fails to launch, and `kit dev stop` does not delete the cached contents, since `llmHarness.Stop()` fails with 

```
Stopping development server...
[ERROR] Failed to stop dev server: no running process found with PID 1842144
```

pkg/cmd/dev/dev.go
```diff
@@ -124,3 +169,69 @@ func findModelFile(absPath string) (string, error) {
...
 	return modelPath, nil
 }
+
+// extractModelKitToCache extracts a ModelKit reference to a cache directory
+// using the unpack library with model filter
+func extractModelKitToCache(ctx context.Context, options *DevStartOptions) error {
+	output.Infof("Extracting ModelKit %s to cache directory...", options.modelRef.String())
+
+	// Use consistent cache directory for extraction
+	extractDir := filepath.Join(options.configHome, "dev-models", "current")
```
amisevsk: We normally put all directories Kit uses as functions in `constants`, in order to keep their definitions in one place

pkg/cmd/dev/dev.go
```diff
@@ -124,3 +169,69 @@ func findModelFile(absPath string) (string, error) {
...
+	}
+	if err := os.Chdir(extractDir); err != nil {
+		return fmt.Errorf("failed to change to cache directory %s: %w", extractDir, err)
+	}
+	// Restore original working directory when done
+	defer func() {
+		if err := os.Chdir(originalWd); err != nil {
+			output.Debugf("Failed to restore working directory: %v", err)
+		}
+	}()
```
amisevsk: This looks like something that should be managed inside the `unpack` library function, given that it takes an `UnpackDir` parameter.

pkg/cmd/dev/dev.go
```diff
@@ -124,3 +169,69 @@ func findModelFile(absPath string) (string, error) {
...
+		}
+		return fmt.Errorf("failed to extract ModelKit: %w", err)
+	}
+
+	// Find the Kitfile in the extracted directory
+	kitfilePath := filepath.Join(extractDir, constants.DefaultKitfileName)
+	if _, err := os.Stat(kitfilePath); err != nil {
+		return fmt.Errorf("kitfile not found in extracted ModelKit: %w", err)
+	}
+	options.modelFile = kitfilePath
```
amisevsk: We have a function `filesystem.FindKitfileInPath(<path>)` to search for Kitfiles

pkg/cmd/dev/dev.go
```diff
@@ -52,10 +74,12 @@ func runDev(ctx context.Context, options *DevStartOptions) error {
 		kitfile.Model.Path = resolvedKitfile.Model.Path
 		kitfile.Model.Parts = append(kitfile.Model.Parts, resolvedKitfile.Model.Parts...)
 	}
+
 	modelAbsPath, _, err := filesystem.VerifySubpath(options.contextDir, kitfile.Model.Path)
 	if err != nil {
 		return err
 	}
+
```
amisevsk, line 82: Should we also do `options.cleanup()` if we're returning one of these errors?

pkg/cmd/dev/opts.go
```diff
@@ -44,28 +49,50 @@ func (opts *DevBaseOptions) complete(ctx context.Context, _ []string) error {
...
+				if opts.modelRef == nil && opts.contextDir == "" {
+					return fmt.Errorf("no directory or ModelKit reference provided - specify a directory path or ModelKit reference (e.g., myrepo/my-model:latest)")
+				}
+				return fmt.Errorf("failed to find Kitfile in directory %s: %w", opts.contextDir, err)
+			}
+			opts.modelFile = foundKitfile
 		}
-		opts.modelFile = foundKitfile
 	}
+
```
amisevsk: This logic reads strangely to me. Is there a case where we _don't_ want to ensure `contextDir` is set appropriately in the first `if` block? I imagine the logic is something like

* If empty, use current dir (`opts.contextDir = ""` -- though this is the implicit default due to zero values)
* else if a directory path, set context dir appropriately (`opts.contextDir = args[0]`)
* else if a reference, set context to the cache directory (`opts.contextDir = constants.ExtractedDevModelPath(opts.configHome)`)

For the second block, the comment suggests were using the original directory-based logic if the `--file` flag is provided, even if `args[0]` is a reference to a ModelKit -- should this be an error instead? Presumably the ModelKit being referenced will come with its own Kitfile. This looks like it's hiding the simplest case (`kit dev start`) behind two conditional checks; it might be simpler to just check for that case first.

pkg/lib/unpack/core.go
```diff
@@ -93,16 +112,16 @@ func runUnpackRecursive(ctx context.Context, opts *unpackOptions, visitedRefs []
...
 		switch mediaType.BaseType {
 		case constants.ModelType:
-			if !shouldUnpackLayer(config.Model, opts.filterConfs) {
+			if !shouldUnpackLayer(config.Model, opts.FilterConfs) {
 				continue
 			}
 			layerInfo = config.Model.LayerInfo
 			layerPath = config.Model.Path
-			output.Infof("Unpacking model %s to %s", config.Model.Name, config.Model.Path)
+			output.Infof("Unpacking model %s to %s", config.Model.Name, opts.UnpackDir)
```
amisevsk: If we want to include the context directory:
```suggestion
			output.Infof("Unpacking model %s to %s", config.Model.Name, filepath.Join(opts.UnpackDir, config.Model.Path))
```

pkg/cmd/dev/dev.go
gorkem: These are all errors on the extracted content so I want to leave the cache directory after post-extraction for debugging. We will cleanup the cache on the next invocation anyway.

pkg/cmd/dev/opts.go
gorkem: I got  less ifs now

PR #978, opened 2025-10-08
Support untagging remote ModelKits

amisevsk:
### Description
The behaviour of `kit remove` when the `--remote` flag is used is unclear; while `kit` distinguishes between "untag" and "delete" for locally-stored ModelKits, for remote operations it always assumes a "delete" operation.

This PR changes behaviour for removing remote ModelKits as follows:

* If the remove command uses a tag (`kit remove --remote <registry>/<org>/<repo>:<tag>`), the ModelKit is only untagged, and not deleted. Other tags pointing to that ModelKit are still valid and it can be pulled by digest
    * Depending on registry settings, untagged ModelKits _may_ be pruned, but `kit` does not do this operation
    * If the remote does not support untagging, an error is returned
* If the remove command uses a digest (`kit remove --remote <registry>/<org>/<repo>@sha256:<digest>`, the ModelKit itself is deleted, removing all tags that point to that digest (i.e. the ModelKit is deleted from the remote, if supported)
* If the remove command uses a tag _and_ the `--force` flag (`kit remove --remote --force <registry>/<org>/<repo>:<tag>`), Kit will resolve the digest and delete the underlying ModelKit, as though a digest was provided for the command.

Note: the oras-go library we are using does not implement the `Untagger` interface on remote repositories; as a result, I've added a small local implementation that we should eventually upstream into the library.

### Testing
I've tested these changes against a locally-running registry container:
```bash
docker run --rm -d \
    --name registry \
    -e REGISTRY_HTTP_ADDR=:5050 \
    -p 5050:5050 \
    registry
kit push <model> localhost:5050/<model> --plain-http
```
This registry supports untagging and deleting artifacts. Kit _should_ be able to handle cases where it's not supported as well, but I haven't tested this directly (as long as the OCI spec is followed).

### Linked issues
I don't think we created an issue for this one

### AI-Assisted Code

### DIFF
--- a/pkg/lib/repo/remote/repository.go
+++ b/pkg/lib/repo/remote/repository.go
@@ -60,6 +60,8 @@ func (r *Repository) Untag(ctx context.Context, reference string) error {
 	switch resp.StatusCode {
 	case http.StatusBadRequest, http.StatusMethodNotAllowed:
 		return fmt.Errorf("remote registry does not support untagging")
+	case http.StatusNotFound:
+		return fmt.Errorf("reference %s not found in remote registry", reference)
 	case http.StatusAccepted:
 		return nil
 	default:
--- a/pkg/cmd/remove/remote.go
+++ b/pkg/cmd/remove/remote.go
@@ -71,5 +71,8 @@ func untagRemoteModel(ctx context.Context, tag string, repo registry.Repository)
 	if !ok {
 		return fmt.Errorf("remote repository implementation does not support untagging ModelKits")
 	}
-	return untaggerRepo.Untag(ctx, tag)
+	if err := untaggerRepo.Untag(ctx, tag); err != nil {
+		return fmt.Errorf("failed to untag ModelKit in remote: %w", err)
+	}
+	return nil
 }


pkg/lib/repo/remote/repository.go
```diff
@@ -42,6 +42,31 @@ type Repository struct {
...
+	}
+	resp, err := r.client().Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to untag: %w", err)
+	}
+	defer resp.Body.Close()
+	switch resp.StatusCode {
+	case http.StatusBadRequest, http.StatusMethodNotAllowed:
+		return fmt.Errorf("remote registry does not support untagging")
+	case http.StatusAccepted:
```
gorkem, line 65: According to [spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints). this endpoint can also return 404.

pkg/lib/repo/remote/repository.go
```diff
@@ -288,3 +313,11 @@ func buildRepositoryBlobUploadURL(plainHTTP bool, ref registry.Reference) string
 	}
 	return fmt.Sprintf("%s://%s/v2/%s/blobs/uploads/", scheme, ref.Host(), ref.Repository)
 }
+
+func buildRepositoryManifestsUrl(plainHTTP bool, registryRef registry.Reference, manifestRef string) string {
```
gorkem: Should Url be URL on the function name, like above?

pkg/lib/repo/remote/repository.go
amisevsk: Good catch, I think I was assuming we had resolved the tag already (which would return the 404 earlier). I'll add a clear message for that case

pkg/lib/repo/remote/repository.go
amisevsk: My least favourite Golang suggestion (use URL instead of Url) :D

Issue #872, opened 2025-05-14
Implement ModelPack specification


gorkem:
**Describe the problem you're trying to solve**
ModelPack defines a specification that is largely compatible with ModelKits. Currently, the Kit CLI cannot consume or manage OCI artifacts that conform to the ModelPack specification, preventing users from leveraging ModelPack-compliant artifacts in our tooling.

**Describe the solution you'd like**
- Extend the Kit CLI to fully support OCI artifacts conforming to the ModelPack specification.
- Ensure that artifacts created with ModelPack remain compliant and are handled correctly by the CLI.
- Enhance Kitfiles so they can generate ModelPack-compliant artifacts directly.

**Describe alternatives you've considered**
- **Separate converter tool:** Translate ModelPack artifacts into ModelKit format, but this adds maintenance overhead and fragments the user workflow.

[ModelPack spec](https://github.com/CloudNativeAI/model-spec)

PR #1000, opened 2025-10-30
Add support for CNCF model-spec artifact (ModelPacks)

amisevsk:
### Description
Add support for the CNCF [model-spec](https://github.com/modelpack/model-spec) format:

* Kit can be used to pack ModelPack artifacts (instead of ModelKits) by supplying the `--use-model-pack` flag to `kit pack`
  * Currently, this does not support the `*+raw` media types in model-spec, nor does it support zstd compression, though the groundwork is there
  * For attaching Kitfiles to ModelPack artifacts, we currently serialize the Kitfile to JSON and attach it as an annotation on the manifest. This matches what `modctl` does for its Modelfile format.
* Unpack should work for both formats. If the artifact does not contain a Kitfile, unpack _should_ still work provided the layers use the `org.cncf.model.filepath` model-spec annotation for filename
* Other commands (e.g. push, pull, inspect, etc.) should work transparently for both artifact types; in some cases a warning will be printed, e.g. if a Kitfile cannot be found.

Part of this PR is a fairly significant refactor of how mediatypes are handled. Mediatypes are now handled through an interface that should make it easy to handle both ModelKits and ModelPack artifacts in a similar way. One fairly significant departure is that Kit can no longer assume every artifact it processes has a Kitfile (as ModelPack artifacts not created by Kit generally won't have one), which introduces a number of additional changes.

Note for testing this PR against `modctl`: current releases of modctl use different mediatypes than the model-spec definition, so some commands may not be compatible. For details, see issue https://github.com/modelpack/modctl/issues/317

### Testing
I've pushed a sample ModelPack artifact as generated by Kit to `jozu.ml/angel/modelpack:kit-generated-modelpack`. This is a simple use case, where we're not packing directories (and especially not nested/intersecting directories) and also hasn't been tested against `modctl` (see above issue)

For testing, you should be able to do all normal KitOps commands with this artifact (i.e. pull, unpack, pack, inspect, etc). To see the manifest, you can use `kit inspect`. To peek at the config in ModelPack format, you can use [crane](https://github.com/google/go-containerregistry/tree/main/cmd/crane):

```bash
crane config jozu.ml/angel/modelpack:kit-generated-modelpack | jq '.'
```

### Linked issues
Closes https://github.com/kitops-ml/kitops/issues/872

### AI-Assisted Code

### DIFF
--- a/pkg/artifact/kitfile.go
+++ b/pkg/artifact/kitfile.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"io"
+	"slices"
 	"time"
 
 	modelspecv1 "github.com/modelpack/model-spec/specs-go/v1"
@@ -191,5 +192,8 @@ func (kf *KitFile) collectLicenses() []string {
 	for _, code := range kf.Code {
 		licenses = appendNotEmpty(licenses, code.License)
 	}
+	slices.Sort(licenses)
+	licenses = slices.Compact(licenses)
+
 	return licenses
 }
--- a/pkg/artifact/kitfile_test.go
+++ b/pkg/artifact/kitfile_test.go
@@ -91,3 +91,109 @@ func loadAllTestCasesOrPanic[T interface{ withName(string) T }](t *testing.T, te
 	}
 	return tests
 }
+
+func TestCollectLicenses(t *testing.T) {
+	testcases := []struct {
+		desc     string
+		kitfile  string
+		licenses []string
+	}{
+		{
+			desc: "Empty kitfile",
+			kitfile: `
+manifestVersion: 1.0.0
+`,
+			licenses: nil,
+		},
+		{
+			desc: "Kitfile with no licenses",
+			kitfile: `
+manifestVersion: 1.0.0
+package:
+  name: "test-package"
+model:
+  path: model-files
+`,
+			licenses: nil,
+		},
+		{
+			desc: "Kitfile with package license",
+			kitfile: `
+manifestVersion: 1.0.0
+package:
+  name: "test-package"
+  license: "Apache-2.0"
+model:
+  path: model-files
+`,
+			licenses: []string{"Apache-2.0"},
+		},
+		{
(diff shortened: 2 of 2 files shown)


pkg/lib/constants/mediatype/modelpack_test.go
```diff
@@ -0,0 +1,82 @@
+package mediatype
+
+// Copyright 2024 The KitOps Authors.
```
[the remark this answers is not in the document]
amisevsk: Don't even know how that happened :thinking:

amisevsk:
Snyk is reporting two issues with the `gopkg.in/yaml.v3` package. These issues are "fixed" in versions `gopkg.in/yaml.v3@3.0.0-20220521103104-8f96da9f5d5e` and `gopkg.in/yaml.v3@3.0.1`, respectively. 

Currently, our `go.mod` uses version `gopkg.in/yaml.v3@v3.0.1`. Suffice it to say: I'm at a loss here. I'm marking the check as successful, as even the reported issues are not a problem for KitOps (malicious yaml input can trigger a crash, which doesn't impact much other than the CLI itself -- you're free to crash your CLI if you like, I suppose).

pkg/artifact/kitfile.go
```diff
@@ -133,3 +136,60 @@ func (kf *KitFile) MarshalToYAML() ([]byte, error) {
...
+	model := modelspecv1.Model{
+		Descriptor: modelDescriptor,
+		ModelFS:    modelFS,
+		Config:     modelConfig,
+	}
+
+	return model
+}
+
+func (kf *KitFile) collectLicenses() []string {
```
gorkem, line 174: Should the licenses be deduplicated?

bergwolf:
Nice! Could you also add a document in https://github.com/modelpack/model-spec/tree/main/docs to explain how to use kitops with modelpack?

pkg/artifact/kitfile.go
amisevsk: Good point; makes sense since licenses aren't tied to individual components in this list.

Issue #1002, opened 2025-10-30
`kit list` for remote repositories does not work since v1.8.0


amisevsk:
**Describe the bug**
v1.8.0 included changes from https://github.com/kitops-ml/kitops/pull/935, which updates OCI reference parsing to add `latest` as the tag whenever it is not present. Since this PR impacts all parsing of references, it breaks commands that do not require a tag in the reference, such as `kit list`

**To Reproduce**
Attempt to `kit list` in a remote repository; normally this is done by specifying a repository name:
```
❯ kit list jozu.ml/angel/modelpack
[ERROR] Failed to resolve reference latest: jozu.ml/angel/modelpack:latest: not found
```
While parsing `jozu.ml/angel/modelpack`, the CLI is implicitly adding the `latest` tag, so instead of listing all ModelKits in that repository it attempts to resolve the `latest` tag, which does not exist.

**Version**
v1.8.0 and up

**Additional context**

PR #1003, opened 2025-10-30
Handle missing tag in ModelKit references per command

amisevsk:
### Description
Explicitly handle cases where a ModelKit reference is missing a tag or digest based on usage. This is required as skipping the tag in a reference is a common error among users, and it helps to clearly report the issue.

For some commands, we explicitly fail the command, requiring a tag, with a clear message (e.g. kit remove). For others, we default to using the tag 'latest', optionally logging an info line to make sure the user is aware.

This PR reverts https://github.com/kitops-ml/kitops/pull/935, as setting `latest` as a default _everywhere_ is not appropriate.

### Linked issues
Fixes https://github.com/kitops-ml/kitops/issues/1002

### AI-Assisted Code

### DIFF
--- a/pkg/cmd/dev/opts_test.go
+++ b/pkg/cmd/dev/opts_test.go
@@ -138,12 +138,6 @@ func TestDevStartOptions_Complete_ErrorCases(t *testing.T) {
 		expectErrorContains string
 	}{
 		{
-			name: "nonexistent directory treated as reference",
-			args: []string{"./nonexistent"},
-			// This will be treated as a ModelKit reference, not a directory
-			// So it won't error in complete(), but would fail later during extraction
-		},
-		{
 			name:                "no arguments with no kitfile in current dir",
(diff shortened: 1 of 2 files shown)


amisevsk:
> We have multiple different versions of essentially the same error message.

I did this intentionally -- inspect/info are read-only commands whereas remove and tag are write commands, and I wanted the message shown to the user to match the importance. Missing a tag in info/inspect isn't a big deal, but if you are attempting to remove a ModelKit (esp. from a remote) you should be careful with the reference. 

For tag, we'll always need different versions of the message, as we need to distinguish between source and target -- using one message for all cases would result in messages like
```
[ERROR] failed to parse source reference: tag or digest is required for ModelKit reference [...]
[ERROR] failed to parse target reference: tag or digest is required for ModelKit reference [...]
```
which IMO are less user-friendly.

PR #1023, opened 2025-11-18
Add cosign-based signing, attestation, and verification for modelkits

### LINKED ISSUE #857: Sign the Vibes
Many users already rely on [Cosign](https://github.com/sigstore/cosign) for signing and attesting OCI artifacts. Today you can use it to sign KitOps ModelKits - but it's a two step / two tool process.

To provide a seamless end-to-end experience within KitOps, we should add analogous `kit sign` and `kit attest` subcommands to the kit CLI. This will allow users to sign ModelKits and attach attestations without switching tools.

New subcommands:
- kit sign
- kit attest
- kit verify

Follow the same flag semantics as cosign for sing and attest but verify should cover both ModelKit signing and attestation verification.

cr3ativ3cod3r:
### Description
This PR adds support for signing, attesting, and verifying modelkits using the kit commands directly. These commands use cosign internally. The users doesn't have to switch between multiple tools. The verify command enables the user to run both verify and verify attestation using a single command.

### Linked issues
closes #857

### AI-Assisted Code

### DIFF
--- a/pkg/cmd/attest/attest.go
+++ b/pkg/cmd/attest/attest.go
@@ -1,3 +1,19 @@
+// Copyright 2025 The KitOps Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
 package attest
 
 import (
--- a/pkg/cmd/attest/attest.go
+++ b/pkg/cmd/attest/attest.go
@@ -1,0 +1,22 @@
+package attest
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+func RunAttest(context context.Context, options *attestOptions) any {
+	cmd := exec.CommandContext(context, "cosign", options.cosignArgs...)
+
+	cmd.Stdin = os.Stdin
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+
+	err := cmd.Run()
+	if err != nil {
+		return fmt.Errorf("attestation failed %s", err)
+	}
+	return nil
+}
--- a/pkg/cmd/attest/attest.go
+++ b/pkg/cmd/attest/attest.go
@@ -23,16 +23,21 @@ import (
 	"os/exec"
 )
 
-func RunAttest(context context.Context, options *attestOptions) any {
+func RunAttest(context context.Context, options *attestOptions) error {
+	_, err := exec.LookPath("cosign")
+	if err != nil {
+		fmt.Println()
+		return fmt.Errorf("cosign not found, please install cosign")
+	}
 	cmd := exec.CommandContext(context, "cosign", options.cosignArgs...)
 
 	cmd.Stdin = os.Stdin
 	cmd.Stderr = os.Stderr
 	cmd.Stdout = os.Stdout
 
-	err := cmd.Run()
+	err = cmd.Run()
 	if err != nil {
-		return fmt.Errorf("attestation failed %s", err)
+		return fmt.Errorf("attestation failed: %w", err)
 	}
 	return nil
 }
--- a/pkg/cmd/attest/cmd.go
+++ b/pkg/cmd/attest/cmd.go
@@ -42,7 +42,7 @@ func (opts *attestOptions) complete(ctx context.Context, args []string) error {
 		return fmt.Errorf("default config path not set on command context")
 	}
 	opts.configHome = configHome
-	opts.cosignArgs = append([]string{"attest"}, args...)
+	opts.cosignArgs = args
 	return nil
 }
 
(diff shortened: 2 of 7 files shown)


pkg/cmd/sign/sign.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+func RunSign(ctx context.Context, options *signOptions) error {
+	cmd := exec.CommandContext(ctx, "cosign", options.cosignArgs...)
```
gorkem, line 33: Commands will panic with "executable file not found" when cosign binary is not installed on the system.  This needs pre-flight checks before making these calls. Also is there a possibility to use cosign as a library instead of an external CLI ?

pkg/cmd/attest/attest.go
```diff
@@ -0,0 +1,38 @@
...
+package attest
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+func RunAttest(context context.Context, options *attestOptions) any {
```
gorkem: Why does this return `any` and not `error`?

pkg/cmd/verify/cmd.go
```diff
@@ -0,0 +1,102 @@
...
+			}
+			return completion.GetLocalModelKitsCompletion(cmd.Context(), toComplete), cobra.ShellCompDirectiveNoFileComp | cobra.ShellCompDirectiveNoSpace
+		},
+		DisableFlagParsing: true,
+	}
+
+	return cmd
+}
+
+func runCommand(opts []verifyOptions) func(cmd *cobra.Command, args []string) error {
```
gorkem: While this code works due to reassignment of `opts` it's confusing. The function signature `runCommand([]verifyOptions{})`suggests it takes an initialized slice, but it's always called with an empty slice and immediately populated. Refactor to use local variable for opts

pkg/cmd/sign/sign.go
```diff
@@ -0,0 +1,38 @@
...
+
+	cmd.Stdin = os.Stdin
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+
+	err := cmd.Run()
+	if err != nil {
+		return fmt.Errorf("signing failed %s", err)
```
gorkem: Should use `%w` instead of `%s`

pkg/cmd/verify/cmd.go
```diff
@@ -0,0 +1,102 @@
...
+	opts.configHome = configHome
+	opts.cosignArgs = args
+
+	return nil
+}
+
+func VerifyCommand() *cobra.Command {
+
+	cmd := &cobra.Command{
+		Use:     "verify [FLAGS]",
```
gorkem: FLAGS -> flags

pkg/cmd/attest/cmd.go
```diff
@@ -0,0 +1,81 @@
...
+		return fmt.Errorf("default config path not set on command context")
+	}
+	opts.configHome = configHome
+	opts.cosignArgs = args
+	return nil
+}
+
+func AttestCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "attest [FLAGS]",
```
gorkem: FLAGS -> flags

pkg/cmd/verify/cmd.go
```diff
@@ -0,0 +1,102 @@
...
+			}
+		}
+
+		for i := range len(opts) {
+			err := RunVerify(cmd.Context(), opts[i])
+			if err != nil {
+				return output.Fatalf("Failed to %s: %s", commands[i], err)
+			}
+		}
+		output.Infof("Modelkit signed")
```
gorkem: Looks like a copy/paste error should be `output.Infof("Modelkit verification successful")`

gorkem (changes requested):

pkg/cmd/attest/attest.go
cr3ativ3cod3r: Changed it to error

pkg/cmd/verify/cmd.go
cr3ativ3cod3r: Made opts local to runCommand instead of passing it as input.

pkg/cmd/sign/sign.go
cr3ativ3cod3r: Changed to %w.

pkg/cmd/sign/sign.go
cr3ativ3cod3r: Looked into use of cosign as a library. Found cosign(github.com/sigstore/cosign/v2/pkg/cosign). But it doesn't have a high level api to sign or attest. Actual signing happens in internal packages. Reviewed sigstore-go, which currently doesn't have attestation creation support.
I have added a check for cosign binary before execution of command.

pkg/cmd/attest/attest.go
```diff
@@ -0,0 +1,43 @@
...
+	cmd.Stdin = os.Stdin
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+
+	err = cmd.Run()
+	if err != nil {
+		return fmt.Errorf("attestation failed: %w", err)
+	}
+	return nil
+}
```
amisevsk, line 43: The implementations for `RunAttest`, `RunSign`, and `RunVerify` appear to be identical.

amisevsk (changes requested):
High level, this PR appears to be a very thin wrapper on just calling the cosign CLI without any additional functionality. As it stands I don't think it's something worth adding to the project. 

Additionally, I haven't built and tested it locally yet, but I have a feeling signing won't work as a user expects -- what happens if the ModelKit has not yet been pushed to a remote repository? 

Lastly, a small nitpick, but adding 3 subcommands to the root command here in my opinion complicates the CLI more than is necessary. They could all be subcommands of one `sign` command to keep the main help page simpler.

amisevsk:
Closing this PR as it has been abandoned.

PR #1029, opened 2025-11-23
Add multi-threading for pack/unpack

### LINKED ISSUE #254: Multi-threading for pack/unpack
The `pack` and `unpack` commands run through layers sequentially, one at a time. This means that if e.g. a modelkit has 3 layers, each 100MB in size, packing/unpacking will take roughly 3x as long as if it was unpacked in parallel. 

Unpack layers in parallel.

N/A

**Additional context**

cr3ativ3cod3r:
### Description
This PR introduces parallel processing of pack and unpack operations. Previously the layers were packed and unpacked sequentially.

- In savekitfilelayers, calls to savecontentlayer is now executed in a new goroutine. They are also indexed, so that the order of layers is preserved.
- In unpackrecursive, calls to unpacklayers is now executed in a new goroutine.
- For the progress bars, created an instance of mpb.Progress and each goroutine adds a bar to this shared instance.
- In ignorepaths struct, added a mutex and used it in matches method as MatchesOrParentMatches is not safe for concurrent access.

### Linked issues
closes #254 

### AI-Assisted Code

### DIFF
--- a/pkg/lib/filesystem/ignore/ignore.go
+++ b/pkg/lib/filesystem/ignore/ignore.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	kfutils "github.com/kitops-ml/kitops/pkg/lib/repo/util"
 
@@ -63,9 +64,12 @@ func New(kitIgnorePaths []string, kitfile *artifact.KitFile, extraLayers ...stri
 type ignorePaths struct {
 	ignoreFileMatcher *patternmatcher.PatternMatcher
 	layers            []string
+	mu                sync.Mutex
 }
 
 func (pm *ignorePaths) Matches(path, layerPath string) (bool, error) {
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
 	path = cleanPath(path)
 	layerPath = cleanPath(layerPath)
 	ignoreFileMatches, err := pm.ignoreFileMatcher.MatchesOrParentMatches(path)


pkg/lib/filesystem/ignore/ignore.go
```diff
@@ -63,9 +64,12 @@ func New(kitIgnorePaths []string, kitfile *artifact.KitFile, extraLayers ...stri
 type ignorePaths struct {
 	ignoreFileMatcher *patternmatcher.PatternMatcher
 	layers            []string
+	mu                sync.Mutex
 }
 
 func (pm *ignorePaths) Matches(path, layerPath string) (bool, error) {
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
```
gorkem, line 72: `saveContentLayer` -> `packLayerToTar` -> `writeLayerToTar`
`writeLayerToTar calls `ignore.Matches` for each file With the mutex, all file traversal becomes serialized. Parallel implementation may be SLOWER than sequential due to lock contention.

pkg/lib/filesystem/local-storage.go
```diff
@@ -136,71 +140,137 @@ func saveConfig(ctx context.Context, localRepo local.LocalRepo, kitfile *artifac
...
+			currentIndex++
+			index := idx
+			path := part.Path
+			eg.Go(func() error {
+				mediaType := mediatype.New(opts.ModelFormat, mediatype.ModelPartBaseType, opts.LayerFormat, opts.Compression)
+				layer, layerInfo, err := saveContentLayer(egCtx, localRepo, path, mediaType, ignore, progress)
+				if err != nil {
+					return err
+				}
+				kitfile.Model.Parts[index].LayerInfo = layerInfo
```
gorkem, line 187: This is introducing concurrent writes to shared Kitfile fields without synchronization, causing potential data races and corruption. The same pattern is repeated on other layers too.

pkg/lib/filesystem/local-storage.go
```diff
@@ -136,71 +140,137 @@ func saveConfig(ctx context.Context, localRepo local.LocalRepo, kitfile *artifac
...
+				return err
+			}
+			kitfile.Docs[index].LayerInfo = layerInfo
+			results <- layerResult{layer: layer, diffID: digest.FromString(layerInfo.DiffId), index: currentIndexCopy}
+			return nil
+		})
+	}
+
+	if err = eg.Wait(); err != nil {
+		close(results)
```
gorkem, line 243: This closes the results channel and returns. However, some goroutines might still be running and will panic when trying to send to the closed channel.

amisevsk:
Closing this PR as it is stale. Feel free to reopen if issues mentioned in review are addressed.

PR #1033, opened 2025-11-28
Add Kubeflow Pipeline components for ModelKits

gorkem:
Introduces push-modelkit and unpack-modelkit components to enable integration of KitOps with Kubeflow Pipelines.

### DIFF
--- a/.github/workflows/kubeflow-components-test.yaml
+++ b/.github/workflows/kubeflow-components-test.yaml
@@ -5,12 +5,6 @@ on:
     paths:
       - 'build/dockerfiles/kubeflow-components/**'
       - '.github/workflows/kubeflow-components-test.yaml'
-  push:
-    branches:
-      - main
-    paths:
-      - 'build/dockerfiles/kubeflow-components/**'
-      - '.github/workflows/kubeflow-components-test.yaml'
 
 jobs:
   test-scripts:
@@ -24,10 +18,6 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y bats
-
-      - name: Install jq
-        run: |
-          sudo apt-get install -y jq
 
       - name: Run push-modelkit tests
         working-directory: build/dockerfiles/kubeflow-components
--- a/build/dockerfiles/kubeflow-components/README.md
+++ b/build/dockerfiles/kubeflow-components/README.md
@@ -31,8 +31,8 @@ If a `Kitfile` exists in `modelkit_dir`, it is used as-is. Otherwise, one is aut
 
 **Outputs**
 
-- `uri` – Tagged ModelKit URI (e.g., `registry.io/myorg/mymodel:v1`)
-- `digest` – Digest-based ModelKit URI (e.g., `registry.io/myorg/mymodel@sha256:abc…`)
+- `ref` – Tagged ModelKit reference (e.g., `registry.io/myorg/mymodel:v1`)
+- `digest` – Digest-based ModelKit reference (e.g., `registry.io/myorg/mymodel@sha256:abc…`)
 
 ### unpack-modelkit
 
@@ -40,7 +40,7 @@ Pulls a ModelKit from a registry and extracts it.
 
 **Inputs**
 
-- `modelkit_uri` – ModelKit reference (e.g., `registry.io/repo:tag` or `registry.io/repo@sha256:…`)
+- `modelkit_reference` – ModelKit reference (e.g., `registry.io/repo:tag` or `registry.io/repo@sha256:…`)
 - `extract_path` – Directory to extract contents (default: `/tmp/model`)
 
 **Outputs**
@@ -89,7 +89,7 @@ def push_modelkit(
     repository: str,
     tag: str,
     input_modelkit_dir: dsl.Input[dsl.Artifact],
-    output_uri: dsl.Output[dsl.Artifact],
+    output_ref: dsl.Output[dsl.Artifact],
     output_digest: dsl.Output[dsl.Artifact],
     modelkit_name: str = '',
     modelkit_desc: str = '',
@@ -111,7 +111,7 @@ def push_modelkit(
             f'--dataset-uri "{dataset_uri}" '
             f'--code-repo "{code_repo}" '
             f'--code-commit "{code_commit}" '
-            f'&& cp /tmp/outputs/uri "{output_uri.path}" '
+            f'&& cp /tmp/outputs/reference "{output_ref.path}" '
             f'&& cp /tmp/outputs/digest "{output_digest.path}"'
         ],
     )
--- a/build/dockerfiles/kubeflow-components/components/push-modelkit/component.yaml
+++ b/build/dockerfiles/kubeflow-components/components/push-modelkit/component.yaml
@@ -17,8 +17,8 @@ inputs:
   - {name: code_commit, type: String, optional: true, description: 'Code commit hash'}
 
 outputs:
-  - {name: uri, type: String, description: 'Tagged ModelKit URI (e.g., registry.io/repo:tag)'}
-  - {name: digest, type: String, description: 'Digest-based ModelKit URI (e.g., registry.io/repo@sha256:...)'}
+  - {name: ref, type: String, description: 'Tagged ModelKit reference (e.g., registry.io/repo:tag)'}
+  - {name: digest, type: String, description: 'Digest-based ModelKit reference(e.g., registry.io/repo@sha256:...)'}
 
 implementation:
   container:
@@ -47,5 +47,5 @@ implementation:
       - {inputValue: code_repo}
       - {inputValue: code_commit}
     fileOutputs:
-      uri: /tmp/outputs/uri
+      ref: /tmp/outputs/reference
       digest: /tmp/outputs/digest
--- a/build/dockerfiles/kubeflow-components/components/unpack-modelkit/component.yaml
+++ b/build/dockerfiles/kubeflow-components/components/unpack-modelkit/component.yaml
@@ -2,7 +2,7 @@ name: unpack-modelkit
 description: Unpack ModelKit artifacts from a registry
 
 inputs:
-  - {name: modelkit_uri, type: String, description: 'ModelKit reference (e.g., registry.io/repo:tag or registry.io/repo@sha256:...)'}
+  - {name: modelkit_reference, type: String, description: 'ModelKit reference (e.g., registry.io/repo:tag or registry.io/repo@sha256:...)'}
   - {name: extract_path, type: String, default: '/tmp/model', description: 'Directory to extract ModelKit artifacts'}
 
 outputs:
@@ -14,7 +14,7 @@ implementation:
     command:
       - /bin/bash
       - /scripts/unpack-modelkit.sh
-      - {inputValue: modelkit_uri}
+      - {inputValue: modelkit_reference}
       - {inputValue: extract_path}
     fileOutputs:
       model_path: /tmp/outputs/model_path
--- a/build/dockerfiles/kubeflow-components/scripts/lib/common.sh
+++ b/build/dockerfiles/kubeflow-components/scripts/lib/common.sh
@@ -69,8 +69,8 @@ die() {
 
 # Retry logic
 retry() {
-    local max_attempts=${1:-3}
-    local delay=${2:-2}
+    local max_attempts=$1
+    local delay=$2
     shift 2
     local attempt=1
 
--- a/build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
+++ b/build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
@@ -91,13 +91,12 @@ if [ ! -d "$MODELKIT_DIR" ]; then
     die "ModelKit directory not found: $MODELKIT_DIR"
 fi
 
-# Construct ModelKit URI
-MODELKIT_URI="${REGISTRY}/${REPOSITORY}:${TAG}"
+# Construct ModelKit reference
+MODELKIT_REF="${REGISTRY}/${REPOSITORY}:${TAG}"
 
-log_info "Starting ModelKit push" "{\"uri\":\"$MODELKIT_URI\"}"
+log_info "Starting ModelKit push" "{\"reference\":\"$MODELKIT_REF\"}"
 
 require_cmd kit cosign jq
-require_env DOCKER_CONFIG
 
 # Disable kit update notifications
 kit version --show-update-notifications=false >/dev/null 2>&1 || true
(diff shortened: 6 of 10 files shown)


.github/workflows/kubeflow-components-test.yaml
```diff
@@ -0,0 +1,61 @@
...
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+
+      - name: Install BATS
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y bats
+
+      - name: Install jq
+        run: |
+          sudo apt-get install -y jq
```
amisevsk: I believe GitHub runners come with jq installed, so this is not necessary

.github/workflows/kubeflow-components-test.yaml
```diff
@@ -0,0 +1,61 @@
...
+  pull_request:
+    paths:
+      - 'build/dockerfiles/kubeflow-components/**'
+      - '.github/workflows/kubeflow-components-test.yaml'
+  push:
+    branches:
+      - main
+    paths:
+      - 'build/dockerfiles/kubeflow-components/**'
+      - '.github/workflows/kubeflow-components-test.yaml'
```
amisevsk: Since these are just tests, should we be running them on pushes to `main`? I don't believe we run any of our other verification actions on push, only on PR.

build/dockerfiles/kubeflow-components/components/push-modelkit/component.yaml
```diff
@@ -0,0 +1,51 @@
...
+  - {name: modelkit_desc, type: String, optional: true, description: 'Description for the ModelKit'}
+  - {name: modelkit_author, type: String, optional: true, description: 'Author for the ModelKit'}
+
+  - {name: dataset_uri, type: String, optional: true, description: 'Dataset URI'}
+  - {name: code_repo, type: String, optional: true, description: 'Code repository URL'}
+  - {name: code_commit, type: String, optional: true, description: 'Code commit hash'}
+
+outputs:
+  - {name: uri, type: String, description: 'Tagged ModelKit URI (e.g., registry.io/repo:tag)'}
+  - {name: digest, type: String, description: 'Digest-based ModelKit URI (e.g., registry.io/repo@sha256:...)'}
```
amisevsk: Nit: the examples here are references, not technically URIs -- references don't contain protocol, but URIs normally would


```suggestion
  - {name: ref, type: String, description: 'Tagged ModelKit reference (e.g., registry.io/repo:tag)'}
  - {name: digest, type: String, description: 'Digest-based ModelKit reference (e.g., registry.io/repo@sha256:...)'}
```

build/dockerfiles/kubeflow-components/components/push-modelkit/component.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - |
+        /scripts/push-modelkit.sh \
+          "$0" "$1" "$2" \
+          --modelkit-dir "$3" \
+          ${4:+--name "$4"} \
+          ${5:+--desc "$5"} \
+          ${6:+--author "$6"} \
+          ${7:+--dataset-uri "$7"} \
+          ${8:+--code-repo "$8"} \
+          ${9:+--code-commit "$9"}
```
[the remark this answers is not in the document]
amisevsk: I think this is bad advice, this will behave how we expect.

build/dockerfiles/kubeflow-components/components/unpack-modelkit/component.yaml
```diff
@@ -0,0 +1,20 @@
+name: unpack-modelkit
+description: Unpack ModelKit artifacts from a registry
+
+inputs:
+  - {name: modelkit_uri, type: String, description: 'ModelKit reference (e.g., registry.io/repo:tag or registry.io/repo@sha256:...)'}
```
amisevsk: Nit: bit more consistency in URI vs reference

```suggestion
  - {name: modelkit_ref, type: String, description: 'ModelKit reference (e.g., registry.io/repo:tag or registry.io/repo@sha256:...)'}
```

build/dockerfiles/kubeflow-components/components/unpack-modelkit/component.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+outputs:
+  - {name: model_path, type: String, description: 'Directory where ModelKit artifacts were extracted'}
+
+implementation:
+  container:
+    image: ghcr.io/kitops-ml/kitops-kubeflow:latest
+    command:
+      - /bin/bash
+      - /scripts/unpack-modelkit.sh
```
amisevsk, line 16: Does this need arguments? Also, should we use `["/bin/bash", "-c"]` for consistency?

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
```diff
@@ -0,0 +1,232 @@
...
+    die "ModelKit directory not found: $MODELKIT_DIR"
+fi
+
+# Construct ModelKit URI
+MODELKIT_URI="${REGISTRY}/${REPOSITORY}:${TAG}"
+
+log_info "Starting ModelKit push" "{\"uri\":\"$MODELKIT_URI\"}"
+
+require_cmd kit cosign jq
+require_env DOCKER_CONFIG
```
amisevsk: Why is `DOCKER_CONFIG` required?

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
```diff
@@ -0,0 +1,232 @@
...
+# Check if Kitfile exists, if not run kit init
+if [ ! -f "$WORK_DIR/Kitfile" ] && [ ! -f "$WORK_DIR/kitfile" ] && [ ! -f "$WORK_DIR/.kitfile" ]; then
+    log_info "No Kitfile found, running kit init"
+
+    INIT_ARGS=()
+    [ -n "$MODELKIT_NAME" ] && INIT_ARGS+=(--name "$MODELKIT_NAME")
+    [ -n "$MODELKIT_DESC" ] && INIT_ARGS+=(--desc "$MODELKIT_DESC")
+    [ -n "$MODELKIT_AUTHOR" ] && INIT_ARGS+=(--author "$MODELKIT_AUTHOR")
+
+    kit init "$WORK_DIR" ${INIT_ARGS[@]+"${INIT_ARGS[@]}"} || die "Failed to initialize Kitfile"
```
amisevsk, line 121: I'm not clear on the purpose of `${INIT_ARGS[@]+"${INIT_ARGS[@]}"}` here

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
```diff
@@ -0,0 +1,232 @@
...
+else
+    log_info "Found existing Kitfile"
+fi
+
+# Pack the ModelKit
+log_info "Packing ModelKit artifacts"
+kit pack "$WORK_DIR" -t "$MODELKIT_URI" || die "Failed to pack ModelKit"
```
amisevsk: Why `|| die `? If the Kit pack fails, the script will exit due to `set -e`.

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
```diff
@@ -0,0 +1,232 @@
...
+
+# Push to registry with retry
+log_info "Pushing to registry"
+retry 3 2 kit push "$MODELKIT_URI" || die "Failed to push ModelKit"
+
+# Extract digest from kit inspect
+log_debug "Extracting digest"
+MODELKIT_DIGEST=$(echo "$MODELKIT_URI" | grep -oE '@sha256:[a-f0-9]+' | sed 's/@sha256://' || echo "")
```
amisevsk: `MODELKIT_URI` is set up as `"${REGISTRY}/${REPOSITORY}:${TAG}"`, which won't contain a digest, unless I'm missing something.

If you want to get a digest, you can use `kit inspect`

```bash
MODELKIT_DIGEST=$(kit inspect "$MODELKIT_URI" | jq -r '.digest')
```

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
```diff
@@ -0,0 +1,232 @@
...
+    set +e
+    INSPECT_OUTPUT=$(kit inspect "$MODELKIT_URI" --remote 2>&1)
+    INSPECT_EXIT_CODE=$?
+    set -e
+
+    log_debug "Kit inspect completed" "{\"exit_code\":$INSPECT_EXIT_CODE}"
+
+    if [ $INSPECT_EXIT_CODE -eq 0 ]; then
+        # Extract digest from JSON output, filtering out any log lines
+        MODELKIT_DIGEST=$(echo "$INSPECT_OUTPUT" | grep -v '^{"timestamp"' | jq -r '.digest' 2>/dev/null | sed 's/sha256://' || echo "")
```
amisevsk: Why filter out `timestamp`? That shouldn't enter the `kit inspect` output as far as I know?

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
```diff
@@ -0,0 +1,232 @@
...
+
+    if [ -z "$MODELKIT_DIGEST" ]; then
+        die "Could not determine ModelKit digest" "{\"reference\":\"$MODELKIT_URI\",\"exit_code\":$INSPECT_EXIT_CODE}"
+    fi
+fi
+
+log_debug "ModelKit digest: $MODELKIT_DIGEST"
+
+# Construct full URI with digest
+FULL_URI="${REGISTRY}/${REPOSITORY}@sha256:${MODELKIT_DIGEST}"
```
amisevsk: Digests read the way you're currently doing it will contain `sha256`, causing this to generate a bad reference:

```
❯ kit inspect st.jozu.ml/angel/test:latest --remote | jq -r '.digest'
sha256:e9cd947e36dae04d923186f03b247f6971ad7bc62957234c2e4e22a73e15d26e
```

build/dockerfiles/kubeflow-components/scripts/lib/common.sh
```diff
@@ -0,0 +1,112 @@
...
+    local extra="${2-}"
+    if [ -z "$extra" ]; then extra="{}"; fi
+    log_error "$1" "$extra"
+    exit 1
+}
+
+# Retry logic
+retry() {
+    local max_attempts=${1:-3}
+    local delay=${2:-2}
```
amisevsk: These vars are treated as unique but there's a positional argument after them -- you'd have to call it `retry "" "" <command>` if you didn't want to provide a max_attempts and delay, which looks a little strange. 

Maybe this should just require those args?

build/dockerfiles/kubeflow-components/components/push-modelkit/component.yaml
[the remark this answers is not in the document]
gorkem: yes it is.

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
gorkem: It is a defensive pattern to handle unbound arrays. In this case it is not going to do much since array is always declared though.

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
gorkem: this way it will also do the json logging.

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
gorkem: copy/paste from workflows. In there it is not done on a subshell so it picks up the `log_json` output sometimes.

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
gorkem: `sed 's/@sha256://'` should strip `sha256:` from the kit inspect output. I believe this works OK.  However this made me realize that it is stripping and re-adding `sha256` for no reason. I will change it to use it sha256.

build/dockerfiles/kubeflow-components/README.md
```diff
@@ -0,0 +1,316 @@
...
+- `ref` – Tagged ModelKit reference (e.g., `registry.io/myorg/mymodel:v1`)
+- `digest` – Digest-based ModelKit reference (e.g., `registry.io/myorg/mymodel@sha256:abc…`)
+
+### unpack-modelkit
+
+Pulls a ModelKit from a registry and extracts it.
+
+**Inputs**
+
+- `modelkit_reference` – ModelKit reference (e.g., `registry.io/repo:tag` or `registry.io/repo@sha256:…`)
```
amisevsk, line 43: Minor nitpick but I just noticed whatever was used to type this replaced `...` with the unicode ellipsis, which looks weird in monospace font: `…`. Here and on line 35

build/dockerfiles/kubeflow-components/README.md
```diff
@@ -0,0 +1,316 @@
...
+            f'"{registry}" "{repository}" "{tag}" '
+            f'--modelkit-dir "{input_modelkit_dir.path}" '
+            f'--name "{modelkit_name}" '
+            f'--desc "{modelkit_desc}" '
+            f'--author "{modelkit_author}" '
+            f'--dataset-uri "{dataset_uri}" '
+            f'--code-repo "{code_repo}" '
+            f'--code-commit "{code_commit}" '
+            f'&& cp /tmp/outputs/reference "{output_ref.path}" '
+            f'&& cp /tmp/outputs/digest "{output_digest.path}"'
```
amisevsk, line 115: If we're required to provide each arg to a shell script, I'm not sure what we gain over just using a Kit CLI container, though maybe I'm missing something.

build/dockerfiles/kubeflow-components/scripts/unpack-modelkit.sh
```diff
@@ -0,0 +1,53 @@
...
+    die "Usage: $0 <modelkit_reference> [extract_path]"
+fi
+
+MODELKIT_REF="$1"
+EXTRACT_PATH="${2:-/tmp/model}"
+
+log_info "Starting unpack" "{\"modelkit_reference\":\"$MODELKIT_REF\",\"extract_path\":\"$EXTRACT_PATH\"}"
+
+require_cmd kit jq
+require_env DOCKER_CONFIG
```
amisevsk, line 22: Why is DOCKER_CONFIG required?

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
```diff
@@ -0,0 +1,225 @@
...
+# Parse arguments
+if [ $# -lt 3 ]; then
+    die "Usage: $0 <registry> <repository> <tag> --modelkit-dir <path> [options]"
+fi
+
+# First three args are positional
+REGISTRY="$1"
+REPOSITORY="$2"
+TAG="$3"
+shift 3
```
amisevsk, line 46: I wonder about the usability of splitting a reference into three components instead of taking a single argument and processing it to get what we want. I.e. why

```bash
./push-modelkit.sh "docker.io" "amisevsk/myrepo" "latest"
```
instead of
```bash
./push-modelkit.sh docker.io/amisevsk/myrepo:latest
```

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
```diff
@@ -0,0 +1,225 @@
...
+if [ ! -d "$MODELKIT_DIR" ]; then
+    die "ModelKit directory not found: $MODELKIT_DIR"
+fi
+
+# Construct ModelKit reference
+MODELKIT_REF="${REGISTRY}/${REPOSITORY}:${TAG}"
+
+log_info "Starting ModelKit push" "{\"reference\":\"$MODELKIT_REF\"}"
+
+require_cmd kit cosign jq
```
amisevsk, line 99: Nit: these should be in reverse order


```suggestion
require_cmd kit cosign jq

log_info "Starting ModelKit push" "{\"reference\":\"$MODELKIT_REF\"}"
```

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
amisevsk: IMO if it's not necessary it makes the script harder to read

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
```diff
@@ -0,0 +1,225 @@
...
+
+    INIT_ARGS=()
+    [ -n "$MODELKIT_NAME" ] && INIT_ARGS+=(--name "$MODELKIT_NAME")
+    [ -n "$MODELKIT_DESC" ] && INIT_ARGS+=(--desc "$MODELKIT_DESC")
+    [ -n "$MODELKIT_AUTHOR" ] && INIT_ARGS+=(--author "$MODELKIT_AUTHOR")
+
+    kit init "$WORK_DIR" ${INIT_ARGS[@]+"${INIT_ARGS[@]}"} || die "Failed to initialize Kitfile"
+else
+    log_info "Found existing Kitfile"
+fi
```
amisevsk, line 124: For usability, it might make sense to note that `--name`, `--desc`, and `--author` are being ignored in this case.

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
gorkem: A cosmetic decision for KFP component inputs. Splitting works easier on the KFP UI. I am not sure if this is a worth change to take for the testing cost.

build/dockerfiles/kubeflow-components/scripts/push-modelkit.sh
gorkem: I would like to know if the script was executed or not, hence the log is the first thing there.

amisevsk (approved):

PR #1058, opened 2026-01-13
Add support for trusting additional certs in Kit commands

amisevsk:
### Description
To handle cases where e.g. a registry is served with a self-signed or otherwise default-untrusted certificate, add flag `--tls-cert` to Kit commands that communicate over the network to supplement the existing `--tls-verify=false` and `--plain-http` flags.

### Linked issues
N/A

### AI-Assisted Code

### DIFF
--- a/pkg/lib/network/auth.go
+++ b/pkg/lib/network/auth.go
@@ -71,6 +71,12 @@ func DefaultClient(opts *options.NetworkOptions) (*auth.Client, error) {
 	transport.TLSClientConfig.InsecureSkipVerify = !opts.TLSVerify
 
 	if len(opts.TLSTrustCertPaths) > 0 {
+		if opts.PlainHTTP {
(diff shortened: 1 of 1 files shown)


pkg/lib/network/auth.go
```diff
@@ -66,13 +69,23 @@ func ClientWithAuth(store credentials.Store, opts *options.NetworkOptions) (*aut
 func DefaultClient(opts *options.NetworkOptions) (*auth.Client, error) {
 	transport := http.DefaultTransport.(*http.Transport).Clone()
 	transport.TLSClientConfig.InsecureSkipVerify = !opts.TLSVerify
+
+	if len(opts.TLSTrustCertPaths) > 0 {
+		certPool, err := getCertsTrust(opts)
```
gorkem, line 80: Should we log a warning if `--tls-verify=false` is also present at the sametime as CA certs?

PR #1066, opened 2026-01-20
docs: add Kubeflow integration guide

TheCoder2010-create:
This guide shows how to pull and unpack a ModelKit inside a **Kubeflow Pipelines** step so downstream steps can use the extracted files.


gorkem:
Closing as we already have this documented for the new kubeflow component

PR #1070, opened 2026-01-23
 Fix `kit login` failure when trace logging is enabled

### LINKED ISSUE #807: Can't use `kit login` with trace debugging enabled
**Describe the bug**
The oras library does some type assertions on its http client interfaces during the login process, which means that the wrapped client we use for trace logging fails to work during login. 

**To Reproduce**
`kit login -vv` will always fail with message "client not supported"

**Version**
Latest

**Workaround**
Login will work without trace logging enabled

**Additional context**
Originally raised in https://github.com/kitops-ml/kitops/issues/786

akagami-harsh:
### Description

- Modified `pkg/output/client.go` to use a `LoggingTransport` instead of wrapping the entire client struct. This ensures the client remains satisfied as an `*auth.Client` for the oras library while still logging requests, because The `oras-go` library's `login` function performs a strict type assertion on the registry client, requiring it to be of type `*auth.Client`. Previously, our trace logging mechanism wrapped this client in a custom `LoggingClient` struct to capture request metrics, which caused this type assertion to fail.

### Linked issues

- Fixes #807 a bug where running `kit login` with trace logging enabled (e.g., `kit login -vv`) would fail with the error `client type not supported`

### AI-Assisted Code

### DIFF
--- a/pkg/output/client.go
+++ b/pkg/output/client.go
@@ -19,26 +19,7 @@ package output
 import (
 	"net/http"
 	"time"
-
-	"oras.land/oras-go/v2/registry/remote"
-	"oras.land/oras-go/v2/registry/remote/auth"
 )
-
-type LoggingClient struct {
-	remote.Client
-}
-
-func (c *LoggingClient) Do(req *http.Request) (*http.Response, error) {
-	start := time.Now()
-	resp, err := c.Client.Do(req)
-	duration := float64(time.Since(start)) / float64(time.Millisecond)
-	if err != nil {
-		SafeLogf(LogLevelTrace, "%s %s -> ERROR -- duration %.2f ms", req.Method, req.URL, duration)
-	} else {
-		SafeLogf(LogLevelTrace, "%s %s -> %d -- duration %.2f ms", req.Method, req.URL, resp.StatusCode, duration)
-	}
-	return resp, err
-}
 
 type LoggingTransport struct {
 	http.RoundTripper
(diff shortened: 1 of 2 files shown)


akagami-harsh:
tested it on local machine

- Before
```shell
harshvir@msi:~/kitops$ go run . login -vv localhost:5000
[DEBUG] Using default config directory: /home/harshvir/.local/share/kitops
Username: test
Password: 
[ERROR] Client type not supported
exit status 1
```
- After
```shell
harshvir@msi:~/kitops$ git checkout fix-login-issue 
Switched to branch 'fix-login-issue'
harshvir@msi:~/kitops$ go run . login -vv localhost:5000
[DEBUG] Using default config directory: /home/harshvir/.local/share/kitops
Username: test
Password: 
[ERROR] Failed to validate the credentials for localhost:5000: Get "https://localhost:5000/v2/": dial tcp 127.0.0.1:5000: connect: connection refused
exit status 1
```

pkg/output/client.go
```diff
@@ -39,13 +40,45 @@ func (c *LoggingClient) Do(req *http.Request) (*http.Response, error) {
...
+			RoundTripper: authClient.Client.Transport,
 		}
+		return authClient
+	}
+
+	return &LoggingClient{
+		Client: c,
 	}
-	return c
 }
```
amisevsk, line 49: If we're switching to wrapping the Transport on the client, we should drop the LoggingClient struct entirely; I don't see the value in supporting two flows, and this approach makes it more confusing as to what's returned. 

Instead, how about

* `output` contains a function `WrapHTTPTransport`:
    ```go
    func WrapHTTPTransport(rt http.RoundTripper) http.RoundTripper {
    	if !logLevel.shouldPrint(LogLevelTrace) {
    		return rt
    	}
    	return &LoggingTransport{rt}
    }
    ```
* Invoked where necessary (i.e. in registry.go) via
    ```go
	authClient.Client.Transport = output.WrapHTTPTransport(authClient.Client.Transport)
	reg.Client = authClient
    ```

Then we can drop the remaining code.

amisevsk (commented):
Good idea @akagami-harsh, this is a nice solution :+1: 

I think we can go simpler on the implementation and drop wrapping clients entirely; whoever creates a client can just wrap their transport as necessary instead of wrapping the client. This avoids the need to do a type assertion and handle specifically authClients (since we're taking in a remote.Client interface but overriding an implementation detail within that interface).

pkg/output/client.go
akagami-harsh: Looks good, implemented https://github.com/kitops-ml/kitops/pull/1070/commits/319edaeea3b9647d8504996e4283471c40e858b2

amisevsk (approved):
Looks good to me! Thanks @akagami-harsh

PR #1071, opened 2026-01-23
Add GitHub downloads badge to README

bupd:
Added GitHub downloads badge to README.

### Description

- Add GitHub downloads badge to README

### AI-Assisted Code

- [x] not ai assisted

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@
 [![LICENSE](https://img.shields.io/badge/License-Apache%202.0-yellow.svg)](./LICENSE)
 [![Discord](https://img.shields.io/discord/1098133460310294528?logo=Discord)](https://discord.gg/Tapeh8agYy)
 [![Twitter](https://img.shields.io/twitter/url/http/shields.io.svg?style=social&label=Twitter)](https://twitter.com/kit_ops)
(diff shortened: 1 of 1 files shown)


amisevsk:
cc: @Jwilliamsr @bmicklea up to you if we want to add a download count badge to our readme or not

bmicklea:
I appreciate the sentiment here, but a lot of our downloads happen via brew and pip (for the SDK) so I think this will be misleading. For example - according to GitHub we only did ~150K downloads, but with brew alone that's over 210K (+40%). That's too big a delta to make the GH only number valuable to people.

I'm closing this as won't fix, but feel free to reopen it if you have a different perspective.

bupd:
@bmicklea - there might be a misunderstanding here. since brew is just basically download artifacts from github - https://github.com/kitops-ml/homebrew-kitops/blob/cc8df3b448809fb668f4ec65955454cd68b9b7be/kitops.rb#L13 - which makes the downloads on brew get counted under github downloads too. 

and the current downloads right now is 220k+ which is including the downloads from brew. so basically all downloads are coming from brew will also be counted as github downloads. if that makes sense. 

Since this is more of a cosmetic change, I agree the decision and its upto you to open / merge this. 

Thanks

PR #1074, opened 2026-01-29
feat: allow kit init to work with remote repositories

### LINKED ISSUE #1055: Allow `kit init` to work with remote repositories

The `kit init` command works to generate a Kitfile from a local directory. This functionality is reused in the `kit import` flow, where the same generation code is used to create a Kitfile for e.g. a remote huggingface repository before downloading any data.

Since the infrastructure for init supports remotes already, it would be useful to be able to generate a Kitfile for a remote without requiring a full import. This would allow more automated uses (e.g. generate a Kitfile, edit it via a script to select files, and then run a kit import with that file). 

`kit init` accepts a URL path and generates a Kitfile for the remote resource if possible.

arnab2001:
### Description
This PR updates the `kit init` command to support generating Kitfiles directly from remote repositories (specifically HuggingFace models and datasets) without needing to clone or import them first. This functionality allows users to inspect and generate a Kitfile for a remote resource, which can then be edited before running an import.

Key changes:
- **Remote Detection**: Updated `kit init` to detect if the input path is a remote HuggingFace repository (URL or `org/repo` format).
- **Repo Parsing**: Implemented `hf.ParseHuggingFaceRepo` to correctly identify and parse references for both models and datasets.
- **API Integration**: Updated `hf.ListFiles` to support listing files for both models and datasets using the correct API endpoints via a new [RepositoryType](cci:2://file:///Volumes/arnab%20ssd/github/kitops/pkg/lib/hf/types.go:18:5-18:23) enum.
- **Refactoring**: Updated `kit import` to utilize the improved [hf](cci:2://file:///Volumes/arnab%20ssd/github/kitops/pkg/lib/hf/list.go:44:0-46:1) library components.
- **Testing**: Added unit tests for remote detection logic, repository URL parsing, and package metadata extraction.

### Linked issues
closes #1055

### AI-Assisted Code

### DIFF
--- a/pkg/cmd/kitinit/cmd.go
+++ b/pkg/cmd/kitinit/cmd.go
@@ -24,7 +24,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"syscall"
 
 	"github.com/kitops-ml/kitops/pkg/artifact"
 	"github.com/kitops-ml/kitops/pkg/lib/constants"
@@ -34,7 +33,6 @@ import (
 	"github.com/kitops-ml/kitops/pkg/output"
 
 	"github.com/spf13/cobra"
-	"golang.org/x/term"
 )
 
 const (
@@ -43,9 +41,9 @@ const (
 a basic Kitfile based on common file formats. Any files whose type (i.e. model,
 dataset, etc.) cannot be determined will be included in a code layer.
 
-For local directories, the generated Kitfile is saved in the target directory.
-For remote repositories (e.g. HuggingFace), the Kitfile is printed to stdout
-or saved to a path specified with --output.
+For local directories, the generated Kitfile is saved in the target directory by
+default. Use --output to specify a different path, or --output=- for stdout.
+For remote repositories (--remote), the Kitfile is printed to stdout by default.
 
 By default the command will prompt for input for a name and description for the Kitfile.`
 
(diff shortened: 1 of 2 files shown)


arnab2001:
Requesting review @gorkem  @amisevsk

arnab2001:
@amisevsk  @bmicklea  Requesting review , Thanks

arnab2001:
@amisevsk , I have done the rebase as you suggested , looking forward to your review.

pkg/cmd/kitinit/cmd.go
```diff
@@ -134,7 +246,27 @@ func (opts *initOptions) complete(ctx context.Context, args []string) error {
...
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return err
+		}
+		if opts.path == "~" {
+			opts.path = home
+		} else {
+			opts.path = filepath.Join(home, opts.path[2:])
+		}
+	}
```
amisevsk: I'm not sure this handling is necessary -- `kit init ~/sub/path` works on the current branch as expected for me. On platforms where `~` is not the home directory (e.g. Windows) I don't think we should interpret `~` specially.

pkg/cmd/kitinit/cmd.go
```diff
@@ -158,6 +290,66 @@ func (opts *initOptions) complete(ctx context.Context, args []string) error {
...
+	if strings.HasPrefix(path, "/") {
+		return true
+	}
+	// Windows-style paths
+	if strings.HasPrefix(path, "\\") || (len(path) >= 2 && path[1] == ':') {
+		return true
+	}
+	return false
+}
```
amisevsk: I think the handling in this function is covered by the `os.Stat` logic in `detectRemoteRepo`. Since this function is only called in `detectRemoteRepo`, we can just drop it and its checks (unless there's some edge case I'm missing).

pkg/cmd/kitinit/cmd.go
```diff
@@ -158,6 +290,66 @@ func (opts *initOptions) complete(ctx context.Context, args []string) error {
...
+	return nil
+}
 
+func (opts *initOptions) completeRemote() error {
+	// For remote repos, only prompt if:
+	// 1. stdin is a terminal (interactive session)
+	// 2. stdout is a terminal (not redirected with >)
+	// 3. no output path specified (output goes to stdout)
+	stdoutIsTerminal := term.IsTerminal(int(syscall.Stdout))
+	if util.IsInteractiveSession() && stdoutIsTerminal && opts.outputPath == "" {
```
amisevsk: Is there a reason we're concerned about prompting if the file is redirected? Comparing to the local case, I'm not sure of the concern here -- if it's not an interactive session should be enough as far as I can tell.

arnab2001:
@amisevsk  Resolved the comments

Issue #887, opened 2025-05-30
Support for storing very large datasets remotely in cheaper blob storage


HJJ256:
**Describe the problem you're trying to solve**
Currently, kitops stores any dataset as artifact in the ModelKit, however, this is not feasible for extremely large datasets, ex. image datasets. OCI registry storage costs can be much higher than blob storage.

**Describe the solution you'd like**
Provide support for remote files and store only the checksums (for example md5 or SHA of images)  for the data files in order to maintain version control.

gorkem:
Thanks for opening this discussion. To help us better understand and address your request, could you clarify a few things?
1. Current cost comparison
    - What are you currently paying (or estimating) for storing datasets as OCI registry layers?
    - How does that compare to your blob-storage costs (e.g., S3, Azure Blob, GCS) for the same data volume?
1.	Desired CLI experience
    - When someone runs kit pull or a similar command, do you expect the CLI to automatically download the remote dataset blobs?
    - Or would it be acceptable for the CLI to only fetch a lightweight “pointer” (e.g., checksums + URLs) and let users retrieve the bulk data themselves (perhaps via a separate step)?

Understanding these points will help us design the right balance between cost savings and usability.


### Foreign Layers as one possible solution

One approach to solve this may be to leverage OCI “foreign layers” (sometimes called “referenced blobs”). Instead of pushing a multi-gigabyte dataset directly into the registry, you upload the dataset to your blob store (for example, S3 or Azure Blob) under a content-addressed key (e.g., sha256:<digest>.tar.gz). Then, when you build your Kitfile, you include a entry like this:

```YAML
datasets:
  - name: my large remote dataset
    digest: sha256:<dataset-hash>
    size: <bytes-of-tar>
    path: "https://my-bucket.s3.amazonaws.com/datasets/sha256-<dataset-hash>.tar.gz"
```

This would result on an OCI manifest such as 

```JSON
"layers": [
  {
    "mediaType": "application/vnd.kitops.modelkit.dataset.v1.tar",
    "digest": "sha256:<dataset-hash>",
    "size": <bytes-of-tar>,
    "urls": [
      "https://my-bucket.s3.amazonaws.com/datasets/sha256-<dataset-hash>.tar.gz"
    ]
  }
]

```

[...]

PR #1085, opened 2026-02-10
Add support for datasets stored remotely in S3 buckets

amisevsk:
### Description
This PR includes the following changes:

* Validation for Kitfiles
  * We now check that `manifestVersion` is `1.0.0` in Kitfiles and print a warning if it is not. ~~This is a potentially **breaking change**, as previously manifestVersion was completely unused~~. Hopefully it does not impact too many people, as this is the value used by `kit init`. Validating `manifestVersion` will allow us to more safely introduce Kitfile changes in the future
  * We now validate paths in the Kitfile to ensure they are as expected (e.g. no URLs, modelkit references only for `model` components, etc.)
* Support for remote datasets, with data stored in an S3 bucket:
  * Kitfile datasets include two new fields:
    ```yaml
    datasets:
    name: my-dataset
    path: some/local/path
    remotePath: s3://<bucket-name>/<bucket-key>
    remoteHash: <ETag for remote object>
    ```
  * Datasets with `remotePath` are not packed into ModelKits; instead, Kit verifies that the remoteHash for that dataset corresponds to the ETag for the bucket object
  * Remote datasets can be unpacked by specifying the `--include-remote` flag for `kit unpack`. If specified, Kit will download files from the S3 bucket, verifying ETag
  * S3 support is provided by the AWS Go SDK (v2) and includes environment support as specified there. In addition, Kit explicitly supports the following environment variables:
    * S3_API_ENDPOINT - API endpoint for S3-compatible services
    * S3_REGION - Region for S3
    * S3_ACCESS_KEY_ID - Access key ID
    * S3_SECRET_KEY - Secret access Key

I've tested these changes against AWS S3 and Cloudflare R2 and was able to pack and unpack ModelKits containing remote datasets.

#### Open questions for the implementation
* We're currently only checking ETags, as they are widely supported; some services also support computing actual digests for contents (or potentially in some cases) -- should we also support better checksums
  * Note that ETags can be used to verify that the object hasn't been changed, but they are not straightforward digests of the content; for example, the ETag for blobs uploaded in multiple parts depend on the size of each part
* I'm not sure if we should incorporate `S3_REGION` into the `s3://` reference somehow, as it's often required (and different datasets may be stored in different regions). I don't see a clean way to include it in the `remotePath` and don't want to overpollute the dataset struct

### Linked issues
Closes https://github.com/kitops-ml/kitops/issues/887

### AI-Assisted Code

### DIFF
--- a/pkg/artifact/kitfile.go
+++ b/pkg/artifact/kitfile.go
@@ -29,6 +29,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/kitops-ml/kitops/pkg/output"
 	modelspecv1 "github.com/modelpack/model-spec/specs-go/v1"
 	"github.com/opencontainers/go-digest"
 	"go.yaml.in/yaml/v3"
@@ -140,16 +141,17 @@ type (
 	}
 )
 
-func (kf *KitFile) LoadModel(kitfileContent io.ReadCloser) error {
+func (kf *KitFile) LoadModel(kitfileContent io.ReadCloser) (err error) {
 	decoder := yaml.NewDecoder(kitfileContent)
 	decoder.KnownFields(true)
 	if err := decoder.Decode(kf); err != nil {
 		return err
 	}
-	if err := kf.Validate(); err != nil {
-		return err
+	warnings, err := kf.Validate()
+	for _, warning := range warnings {
+		output.Logln(output.LogLevelWarn, warning)
 	}
-	return nil
+	return err
 }
 
 func (kf *KitFile) MarshalToJSON() ([]byte, error) {
@@ -171,14 +173,15 @@ func (kf *KitFile) MarshalToYAML() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
-func (kf *KitFile) Validate() error {
+func (kf *KitFile) Validate() (warnings []string, err error) {
+	if kf.ManifestVersion != "1.0.0" {
+		warnings = append(warnings, fmt.Sprintf("Unrecognized manifestVersion %s: treating Kitfile as 1.0.0", kf.ManifestVersion))
+	}
+
 	var errs []string
 	addErr := func(format string, a ...any) {
 		s := fmt.Sprintf(format, a...)
 		errs = append(errs, fmt.Sprintf("  * %s", s))
-	}
-	if kf.ManifestVersion != "1.0.0" {
-		addErr("invalid manifestVersion: expect 1.0.0 but got %s", kf.ManifestVersion)
 	}
 
 	// Map of paths to the component that uses them; used to detect duplicate paths
@@ -287,10 +290,10 @@ func (kf *KitFile) Validate() error {
 	if len(errs) > 0 {
 		// Iterating through the paths map is random; sort to get a consistent message
 		slices.Sort(errs)
-		return fmt.Errorf("errors while validating Kitfile: \n%s", strings.Join(errs, "\n"))
+		return warnings, fmt.Errorf("errors while validating Kitfile: \n%s", strings.Join(errs, "\n"))
 	}
 
-	return nil
+	return warnings, nil
 }
 
 func (kf *KitFile) ToModelPackConfig(diffIDs []digest.Digest) modelspecv1.Model {
--- a/pkg/artifact/testdata/validation/bad-manifest-version.yaml
+++ b/pkg/artifact/testdata/validation/bad-manifest-version.yaml
@@ -1,3 +1,0 @@
-kitfile: |
-  manifestVersion: v99.99.99
-errRegexp: "invalid manifestVersion"


pkg/artifact/kitfile.go
```diff
@@ -145,6 +159,81 @@ func (kf *KitFile) MarshalToYAML() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+func (kf *KitFile) Validate() error {
```
gorkem: I think this function is not called from anywhere and I can pack a kitfile with an invalide manifestVersion. Is this expected?

pkg/lib/filesystem/unpack/core.go
```diff
@@ -174,8 +175,17 @@ func unpackRecursive(ctx context.Context, opts *UnpackOptions, visitedRefs []str
...
+			for idx := datasetIdx; idx < len(config.DataSets); idx++ {
+				dataset := config.DataSets[idx]
+				if dataset.RemotePath != "" {
+					continue
+				}
+				entry = dataset
+				datasetIdx = idx + 1
+				break
+			}
 			if !shouldUnpackLayer(entry, opts.FilterConfs) {
```
gorkem: If all datasets are remote. we would call `shouldUnpackLayer` with zero-value entry. I think we should update `shouldUnpackLayer` to explicitly handle empty entry (not because it has empty Name and Path) or not call shouldUnpackLayer.

pkg/lib/external/s3api/api.go
```diff
@@ -0,0 +1,163 @@
...
+	if region := os.Getenv(RegionEnvVar); region != "" {
+		output.Logf(output.LogLevelTrace, "Using region %s", region)
+		cfgOpts = append(cfgOpts, s3config.WithRegion(region))
+	}
+	if endpoint := os.Getenv(APIEndpointEnvVar); endpoint != "" {
+		output.Logf(output.LogLevelTrace, "Using API endpoint %s", endpoint)
+		clientOpts = append(clientOpts, func(o *s3.Options) { o.BaseEndpoint = aws.String(endpoint) })
+		// For now, use path style URLs (endpoint.com/bucket/key) instead of virtual-hosted style (bucket.endpoint.com/key) when
+		// an alternate endpoint is specified; this is used/supported by most S3-compatible APIs
+		clientOpts = append(clientOpts, func(o *s3.Options) { o.UsePathStyle = true })
```
gorkem, line 153: Appearently use ARN region setting enables cross-region for the bucket urls too. 

```suggestion
		clientOpts = append(clientOpts, func(o *s3.Options) {
            o.UseARNRegion = true
        })
```

gorkem:
if the use ARN config solves the cross region issue for AWS S3, I do not think we need to add region to s3 url somehow. I think this region problem is on AWS only the other compatible may not even have "regions".

pkg/artifact/kitfile.go
amisevsk: Woops, forgot to actually add those calls.

amisevsk: Fixed

pkg/lib/filesystem/unpack/core.go
amisevsk: This should never happen, since the entry into this block is when we encounter a dataset layer in the manifest (i.e. the Kitfile has at least one non-remote dataset). I'll add a guard against this regardless.

pkg/lib/external/s3api/api.go
amisevsk: Lets try it, I'm not sure how all these settings play together.

pkg/artifact/kitfile.go
```diff
@@ -204,3 +293,24 @@ func (kf *KitFile) collectLicenses() []string {
...
+
+	switch parsed.Scheme {
+	case "s3":
+		return S3PathType, nil
+	case "http", "https":
+		return UnknownPathType, fmt.Errorf("HTTP urls are not supported in paths (%s)", path)
+	}
+
+	return UnknownPathType, fmt.Errorf("unrecognized URL in path: %s", path)
+}
```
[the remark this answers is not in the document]
amisevsk: I'm considering leaving this as-is; our path requirements are a little shaky at the moment and I believe we assume they're defined unix-style even on Windows (where paths are converted to Windows-style paths before accessing disk). I don't have access to a Windows machine anymore, so I can't easily test these flows.

In any case, Kitfile paths must be relative, so drive-letter paths are not applicable.

pkg/lib/external/s3api/api.go
```diff
@@ -0,0 +1,174 @@
...
+
+// S3ClientAPI is a wrapper over s3.Client to allow mocking intests
+type S3ClientAPI interface {
+	HeadObject(context.Context, *s3.HeadObjectInput, ...func(*s3.Options)) (*s3.HeadObjectOutput, error)
+	GetObject(context.Context, *s3.GetObjectInput, ...func(*s3.Options)) (*s3.GetObjectOutput, error)
+}
+
+// ParseS3ObjectReference parses a Kit-spec 's3://' reference and hash into an S3ObjectReference. For filling
+// endpoint and region, environment variables are read.
+// The format for an s3:// reference is
```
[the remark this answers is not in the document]
amisevsk: The AWS dependency reads those environment variables, this docline is just noting that.

pkg/lib/external/s3api/api.go
```diff
@@ -0,0 +1,174 @@
...
+		Key:       &ref.Key,
+		VersionId: stringPtrOrNil(ref.Version),
+		IfMatch:   &ref.Hash,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to get object from S3 bucket: %w", err)
+	}
+	defer obj.Body.Close()
+
+	outfile, err := os.Create(outputPath)
```
gorkem, line 116: Noticed that this works different than `saveContentLayer` which writes to a temp file, then `os.Rename` to outputPath.

pkg/artifact/kitfile.go
```diff
@@ -145,6 +162,85 @@ func (kf *KitFile) MarshalToYAML() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+func (kf *KitFile) Validate() error {
+	if kf.ManifestVersion != "1.0.0" {
```
gorkem, line 177: This hard-fail is going to break a lot of existing kitfiles especially  hand-written Kitfiles not generated via kit init. Should we instead try a more gentle approach and print a warning when ManifestVersion is empty?

amisevsk: You're right. I'll downgrade manifestVersion issues to a warning

pkg/lib/external/s3api/api.go
amisevsk: I think the equivalent for this function would be `unpackLayer` -- `DownloadObject` is called only when we want a local copy of the blob from S3. During pack (where we use `saveContentLayer`) we only verify the object exists and its ETag matches the `remoteHash`.

amisevsk (commented):
For some reason, GitHub is interpreting my responses here as part of a review and not letting me submit them separately (unless I delete all of them, I assume)

amisevsk:
I've downgraded manifestVersion errors to (for now) printed warnings. I don't love the idea of including the output package in artifact but this was the cleanest way to add warnings without putting them throughout the repository

I've also remembered we _did_ have some Kitfile validation code elsewhere; I've merged it into the current validate method (see commit 0ec8f95). In this process, I also included a bugfix for https://github.com/kitops-ml/kitops/issues/1109.

pkg/lib/filesystem/unpack/core.go
```diff
@@ -217,6 +230,59 @@ func unpackRecursive(ctx context.Context, opts *UnpackOptions, visitedRefs []str
...
+		if err != nil {
+			return fmt.Errorf("failed to parse S3 object reference for dataset %s: %w", dataset.Path, err)
+		}
+		remoteFiles[dataset.Path] = *ref
+	}
+	if len(remoteFiles) > 0 {
+		if !opts.IncludeRemote {
+			output.Logf(output.LogLevelWarn, "ModelKit contains remote datasets. To unpack, specify the --include-remote flag")
+		} else {
+			client, err := s3api.SetUpClient(ctx)
```
gorkem, line 250: Should this call ` shouldUnpackLayer()` at some point to check if --filter is set?

pkg/artifact/kitfile.go
```diff
@@ -146,6 +173,129 @@ func (kf *KitFile) MarshalToYAML() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+func (kf *KitFile) Validate() (err error, warnings []string) {
```
gorkem: Should this return error as the last value to satisfy the Golang gods.

amisevsk: Yeah, good catch -- I forgot about the Golang gods

pkg/lib/filesystem/unpack/core.go
amisevsk: Good point. I forgot this because originally `path:` contained the remote ref, but now that they're separate fields that makes sense.

PR #1092, opened 2026-02-25
Enhance SKILL.md skill support in Kitfile generation

gorkem:
Treat directories with SKILL.md as prompt skills and prompt metadata comes from frontmatter instead of filename-based prompt detection.

### Changes 

- Detect directories containing SKILL.md and treat them as skill prompts.
- Parse frontmatter fields (name, description, license) from SKILL.md.
- Apply frontmatter metadata to generated prompt entries and package metadata (without overriding explicit user package values).
- Give root-level SKILL.md precedence and generate a single prompt with path ..
- Update prompt unpack filtering to match prompt name with path fallback.
- Add tests

### DIFF
--- a/pkg/lib/kitfile/generate/generate.go
+++ b/pkg/lib/kitfile/generate/generate.go
@@ -94,13 +94,12 @@ func GenerateKitfile(dir *DirectoryListing, packageOpt *artifact.Package) (*arti
 		kitfile.Package = *packageOpt
 	}
 
-	//  SKILL.md: found at ROOT treat entire directory as a skill
-	if found, skillPath := dirContainsSkillMD(*dir); found {
+	// SKILL.md at root: treat entire directory as a single skill
+	if found, _ := dirContainsSkillMD(*dir); found {
 		output.Logf(output.LogLevelTrace, "SKILL.md found; treating as a skill directory")
-		prompt := artifact.Prompt{Path: "."}
-		if fm := parseSkillFrontmatter(skillPath); fm != nil {
-			prompt.Name = fm.Name
-			prompt.Description = fm.Description
+		prompt, fm := buildPromptFromSkill(*dir)
+		kitfile.Prompts = append(kitfile.Prompts, prompt)
+		if fm != nil {
 			if kitfile.Package.Name == "" {
 				kitfile.Package.Name = fm.Name
 			}
@@ -111,7 +110,6 @@ func GenerateKitfile(dir *DirectoryListing, packageOpt *artifact.Package) (*arti
 				kitfile.Package.License = fm.License
 			}
 		}
-		kitfile.Prompts = append(kitfile.Prompts, prompt)
 		return kitfile, nil
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/lib/kitfile/generate/generate.go
```diff
@@ -94,6 +94,27 @@ func GenerateKitfile(dir *DirectoryListing, packageOpt *artifact.Package) (*arti
...
+			if kitfile.Package.Description == "" {
+				kitfile.Package.Description = fm.Description
+			}
+			if kitfile.Package.License == "" {
+				kitfile.Package.License = fm.License
+			}
+		}
+		kitfile.Prompts = append(kitfile.Prompts, prompt)
+		return kitfile, nil
+	}
```
amisevsk: This section appears to be a duplication of `buildPromptFromSkill` and `applySkillMetadataToPackage` -- can we just use those functions instead?

pkg/lib/kitfile/generate/skill.go
```diff
@@ -0,0 +1,128 @@
...
+		if kitfile.Package.Description == "" {
+			kitfile.Package.Description = fm.Description
+		}
+	}
+
+	first := skillFrontmatters[0]
+	if kitfile.Package.License == "" && first.License != "" {
+		kitfile.Package.License = first.License
+		output.Logf(output.LogLevelTrace, "Using license %q from skill %q", first.License, first.Name)
+	}
```
amisevsk: Hm, promoting skill details to the top-level package seems fragile. This is only called at the very end of the normal generation flow (excluding the short-cut "everything here is part of the skill" flow). Does it make sense to promote this information to the Kitfile's package if the Kitfile contains other content?

pkg/lib/kitfile/generate/generate.go
```diff
@@ -211,10 +232,19 @@ func GenerateKitfile(dir *DirectoryListing, packageOpt *artifact.Package) (*arti
 		kitfile.Package.License = detectedLicenseType
 	}
 
+	applySkillMetadataToPackage(kitfile, *dir)
+
```
amisevsk, line 234: Will this call ever have an effect? `dir` in this context is the directory used for `kit init` -- if it contains a `skill.md`, the block at the top

> ```
> //  SKILL.md: found at ROOT treat entire directory as a skill
> if found, skillPath := dirContainsSkillMD(*dir); found {
> ```

will early return on line 115.

pkg/lib/kitfile/generate/skill.go
gorkem: Yeah, it is a bug. This should promote it only if it is a prompt only package.

pkg/lib/kitfile/generate/generate.go
gorkem: Yeap, when root has a SKILL.md this is unreachable — the early return on line 115 handles that case. This call covers the multi-skill layout where the root has no SKILL.md but immediate subdirectories do (e.g. `docx/SKILL.md`, `xlsx/SKILL.md`). Each subdir is added as a prompt by addDirToKitfile, and applySkillMetadataToPackage then promotes license to the package.

pkg/lib/kitfile/generate/generate.go
gorkem: this code was doing some weirdness before I remembered the CWD so it was not able to reuse the functions but now it can.

PR #1094, opened 2026-02-26
feat(list): add --filter support to kit list using unpack filter semantics

### LINKED ISSUE #1091: Add filtering to `kit list`
## Problem

`kit list` currently returns every ModelKit in local storage (or every tag in a remote repository) with no way to narrow results. As ModelKit usage expands beyond ML models to include agent skills, prompt libraries, datasets, and code packages, the unfiltered flat list becomes increasingly hard to work with.

The only "filtering" today is targeting a specific remote repository as a positional argument. There is no way to filter by what a ModelKit contains.

## Proposed Solution

Add a `--filter` / `-f` flag to `kit list` using the same syntax and semantics as the existing `kit unpack --filter` flag.

### Filter syntax

```
--filter <type>[,<type>][:<name>[,<name>]]
```

Where `<type>` is one of the valid Kitfile content types (`model`, `datasets`, `code`, `docs`, `prompts`) and `<name>` is an optional comma-separated list matched against the layer's name or path.

This is identical to `kit unpack --filter`. The same `ParseFilter` function and `FilterConf` struct from `pkg/lib/filesystem/unpack/filter.go` should be reused (or extracted to a shared location).

### Behavior

- `-f` can be specified multiple times; a ModelKit is listed if it matches **any** filter (OR logic across filters, consistent with `kit unpack`)
- Within a single filter, type and name conditions are AND-ed (consistent with `kit unpack`)
- If no filters are specified, all ModelKits are listed (current default behavior)
- A ModelKit matches a filter if its Kitfile contains at least one layer matching the filter
- Works for both local and remote listings
- Compatible with all output formats (`--format table`, `--format json`, Go templates)

### Example usage

```bash
# List only ModelKits that contain prompt layers
kit list -f prompts

# List only ModelKits containing a model
kit list -f model

# List ModelKits that have either prompts or datasets
kit list -f prompts -f datasets

# List ModelKits with a specific named prompt
kit list -f prompts:pdf-processing

```

### Implementation notes

The core filter parsing and matching logic already exists in `pkg/lib/filesystem/unpack/filter.go`.

rishi-jat:
### Description

This PR adds filtering support to `kit list` using the same syntax and semantics as the existing `kit unpack --filter` functionality.

Currently, `kit list` returns all ModelKits without a way to narrow results based on contents. This change introduces a `--filter` / `-f` flag that allows users to list only ModelKits containing specific Kitfile layer types or names.

The implementation strictly reuses the existing filtering logic from `pkg/lib/filesystem/unpack/filter.go` to ensure identical behavior with `kit unpack`.

Key points:

- Adds `--filter` (`-f`) flag to `kit list`
- Supports multiple filters with OR logic across flags
- Applies AND logic within a single filter (type + name)
- Works for both local and remote listings
- Compatible with all output formats (`table`, `json`, Go templates)
- Default behavior remains unchanged when no filters are provided
- No filtering logic duplicated; existing parsing and matching semantics are reused

Example usage:

```
kit list -f prompts
kit list -f model
kit list -f prompts -f datasets
kit list -f prompts:pdf-processing

```

Filtering is applied before output formatting so existing formatters continue to work without modification.

### Linked issues

Closes #1091


rishi-jat:
I’ve manually reviewed the implementation and verified it builds and tests successfully locally.

rishi-jat:
Hey @gorkem , please review the PR when you get a chance. Thanks!

pkg/lib/filesystem/unpack/filter.go
```diff
@@ -126,6 +126,65 @@ func matchesFilters(baseType, field string, filterConfs []FilterConf) bool {
 	return false
 }
 
+// KitfileContainsMatchingLayer returns true if the given Kitfile contains at least one layer
+// that matches any of the provided filters. If filters is empty, returns true.
+// This reuses shouldUnpackLayer to ensure identical matching semantics.
+func KitfileContainsMatchingLayer(kf *artifact.KitFile, filters []FilterConf) bool {
```
gorkem: I will let @amisevsk say if this is the correct place for this new function

gorkem:
overall lgtm. @amisevsk should also take a look.

pkg/lib/filesystem/unpack/filter.go
rishi-jat: okay

rishi-jat:
Hey @amisevsk , This PR needs your review please have a look when you get a chance. Thanks!

pkg/cmd/list/cmd.go
```diff
@@ -113,6 +138,15 @@ func (opts *listOptions) complete(ctx context.Context, args []string) error {
 		opts.format = "template"
 	}
 
+	// Parse filters using library functionality
+	for _, filter := range opts.filters {
+		filterConf, err := unpack.ParseFilter(filter)
```
amisevsk: I think we should avoid cross dependencies in `cmd` package (i.e. having `list` depend on `unpack`). Could we move the filter logic from unpack into a package somewhere within `pkg/lib`? This would allow it to be reused more cleanly in both unpack and list.

pkg/lib/filesystem/unpack/filter.go
amisevsk: I think this is a fine place for the function, but do think that the filter definitions and functions should be moved under `pkg/lib`.

pkg/cmd/list/cmd.go
rishi-jat: okay

rishi-jat:
> Looks good to me generally. I think it'd be better to move filter logic into `pkg/lib` -- perhaps under the `kitfile` package to avoid a dependency from `pkg/cmd/list` on `pkg/cmd/unpack`.

will do the suggested changes. asap

rishi-jat:
@gorkem @amisevsk Requested changes has been made please have a look when you get a chance. Thanks!

PR #1097, opened 2026-02-26
refactor(kit): Expose core Kit CLI functionality as a reusable Go library #905

### LINKED ISSUE #905: Expose core Kit CLI functionality as a reusable Go library
All Kit CLI commands currently live under `cmd/*`, with business logic and flag parsing tightly coupled to Cobra. This makes it difficult for other Go programs to import and invoke Kit functionality without shelling out.

1. Create a new package `pkg/kit` exposing public APIs like:   
   ```go
   func List(ctx context.Context, opts ListOptions) ([]ModelInfo, error)
   func Push(ctx context.Context, opts PushOptions) error
   // …etc…
   ```
1. Refactor each command  to delegate to these package functions
1. Move data types, validation routines, output-format helpers, and error definitions into pkg/kit. 
1.	Update tests to separate unit tests against pkg/kit APIs in  and retain or convert integration tests under cmd/ to ensure the CLI remains unchanged.
1. Add documentation with examples for library usage

arnab2001:
### Description
Refactors CLI command implementations into a reusable Go API under `pkg/kit`.

- Adds `pkg/kit` operations for `Info`, `Inspect`, `List`, `Login`, `Logout`, `Pull`, `Push`, `Remove`, `Tag`, and `Unpack` wrappers.
- Updates CLI command entrypoints (`pkg/cmd/*/cmd.go`) to call `pkg/kit` instead of local command-specific logic.
- Removes migrated legacy command implementation files from `pkg/cmd/*`.
- Moves list helper tests to `pkg/kit/list_test.go`.
- Adds package docs (`pkg/kit/doc.go`, `pkg/kit/README.md`).
- Updates `.gitignore` to ignore root-only binaries (`/kit`, `/kit.exe`).
- Validation: `go test ./...` passes.

### AI-Assisted Code

### DIFF
--- a/pkg/cmd/info/cmd.go
+++ b/pkg/cmd/info/cmd.go
@@ -111,13 +111,13 @@ func runCommand(opts *infoOptions) func(*cobra.Command, []string) error {
 			if err != nil {
 				return output.Fatalln(err)
 			}
-			fmt.Print(string(filteredOutput))
+			fmt.Fprint(cmd.OutOrStdout(), string(filteredOutput))
 		} else {
 			yamlBytes, err := config.MarshalToYAML()
 			if err != nil {
 				return output.Fatalf("Error formatting manifest: %w", err)
 			}
-			fmt.Print(string(yamlBytes))
+			fmt.Fprint(cmd.OutOrStdout(), string(yamlBytes))
 		}
 
 		return nil
--- a/pkg/kit/pack.go
+++ b/pkg/kit/pack.go
@@ -49,6 +49,10 @@ type PackResult struct {
 	Descriptor ocispec.Descriptor
 }
 
+// Pack packs a ModelKit from a Kitfile and context directory into local storage.
+// Note: Pack temporarily changes the process working directory to ContextDir
+// (required for correct tar relative paths) and restores it on return.
+// It is not safe to call concurrently from multiple goroutines.
 func Pack(ctx context.Context, opts *PackOptions) (*PackResult, error) {
 	contextDir, err := filepath.Abs(opts.ContextDir)
 	if err != nil {


arnab2001:
Addressed both Codex P1 suggestions by adding safe default normalization in pkg/kit API entrypoints (pull/login), and re-ran targeted tests for kit/network/repo packages.

pkg/cmd/remove/remove.go
```diff
@@ -66,17 +91,15 @@ func removeAllModels(ctx context.Context, opts *removeOptions) error {
 				output.Errorf("Failed to remove %s@%s: %s", repository, manifestDesc.Digest, err)
 				continue
 			}
-			// Skip future manifest descriptors with this digest, since we just removed it.
 			skipManifests[manifestDesc.Digest] = true
 			output.Infof("Removed %s@%s", repository, manifestDesc.Digest)
 		}
```
amisevsk, line 72: Why have all the comments on this file been removed?

pkg/cmd/remove/cmd.go
```diff
@@ -164,20 +165,16 @@ func runCommand(opts *removeOptions) func(*cobra.Command, []string) error {
...
-		case opts.removeAll && opts.forceDelete:
-			err = removeAllModels(cmd.Context(), opts)
+		kitOpts := &kit.RemoveOptions{
+			NetworkOptions: opts.NetworkOptions,
+			ConfigHome:     opts.configHome,
+			ForceDelete:    opts.forceDelete,
+			RemoveAll:      opts.removeAll,
+			Remote:         opts.remote,
+			ModelRef:       opts.modelRef,
+			ExtraTags:      opts.extraTags,
```
amisevsk: We're now duplicating the options for every command twice -- once in the cmd package and once in the kit package. Why not import the e.g. `kit.RemoveOptions` struct here and bind flags into it instead of defining it twice?

pkg/kit/pull.go
```diff
@@ -86,21 +99,21 @@ func pullParents(ctx context.Context, localRepo local.LocalRepo, desc ocispec.De
...
 	if err != nil {
 		return ocispec.DescriptorEmptyJSON, fmt.Errorf("failed to read repository: %w", err)
 	}
-	if err := referenceIsModel(ctx, opts.modelRef, repo); err != nil {
+	if err := referenceIsModel(ctx, opts.ModelRef, repo); err != nil {
 		return ocispec.DescriptorEmptyJSON, err
 	}
 
-	desc, err := localRepo.PullModel(ctx, repo, *opts.modelRef, &opts.NetworkOptions)
+	desc, err := localRepo.PullModel(ctx, repo, *opts.ModelRef, &opts.NetworkOptions)
```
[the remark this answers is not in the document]
arnab2001: Fixed in 95e4e52 — `kit.Pull` now initializes `Concurrency` to 5 when the caller leaves it at zero, and sets `CredentialsPath` + `TLSVerify` defaults. Same guard was also added to `kit.Push` in 022577d for consistency.

pkg/kit/auth.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// Login logs in to a registry and saves credentials.
+func Login(ctx context.Context, opts *LoginOptions) error {
+	credentialsStorePath := constants.CredentialsPath(opts.ConfigHome)
+	store, err := network.NewCredentialStore(credentialsStorePath)
+	if err != nil {
+		return err
+	}
+	reg, err := remote.NewRegistry(opts.Registry, &opts.NetworkOptions)
```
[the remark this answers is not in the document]
arnab2001: Fixed in 95e4e52 — `kit.Login` now defaults `TLSVerify` to `true` when `CredentialsPath` is empty (same pattern as `kit.Pull`). The `kit.Push` entrypoint also got the same guard in 022577d.

pkg/cmd/remove/remove.go
arnab2001: the comments were accidentally stripped during the initial refactor. Restored all comments in `pkg/kit/remove.go` (the new home for this logic) in 95e4e52, including the skipManifests explanation, untag-vs-delete rationale, and the oras Untagger interface note.

pkg/cmd/remove/cmd.go
arnab2001: Agreed, this was unnecessary duplication. Fixed in 95e4e52 — all cmd packages now embed `kit.*Options` directly (e.g. `type removeOptions struct { kit.RemoveOptions }`) and pass `&opts.RemoveOptions` to the kit function. CLI-specific fields like `format`/`template` in list or `username`/`password` in login are the only local additions. No more manual field-by-field copy.

arnab2001:
## Known limitations & follow-up work

### Not yet wrapped in `pkg/kit`
- **`kitimport`** — git/HuggingFace import (external deps)
- **`kitcache`** — cache management (easy)
- **`version`** — version string (trivial)
- **`dev`** — local LLM server (interactive/long-running, poor fit for library API — intentionally deferred)

These can be follow-up PRs without touching the current refactor, @amisevsk  @gorkem let me know what do you think. 

### Library-layer side effects
Several `pkg/kit` functions call `output.Infof/Debugf` for progress logging (e.g., `Login` prints "Log in successful", `Remove` logs each deleted manifest). This matches the existing CLI behavior and uses the project's `pkg/output` logging layer, which consumers can redirect via `output.SetOut()` / `output.SetLogLevel()`. A future improvement could return structured results instead.

### `kit.Pack` is not goroutine-safe
`Pack` temporarily changes the process CWD (required for correct tar relative paths) and restores it via `defer`. This matches the original CLI behavior but means concurrent `Pack` calls will race. Documented in the function's godoc.

arnab2001:
@amisevsk Let me know of your thoughts if dev, kitimport, kitcache and version should be included in this PR. also waiting for your review 
Thanks.

PR #1098, opened 2026-02-26
Enable loading of lora-adapter parts in dev mode

### LINKED ISSUE #306: enable lora-adapters for dev mode
dev mode should be able to load lora-adapters if they are present in the modelkit

dev mode should detect if there is a lora-adapter type on the modelkit and load it to the dev-mode.

rishi-jat:
### Description

This PR enables dev mode to load LoRA adapters when `type: lora-adapter` parts are defined in the ModelKit.

Dev mode now:

- Detects `lora-adapter` entries in `model.parts`
- Resolves adapter paths within the context directory
- Locates the corresponding `.gguf` adapter file
- Passes `--lora <adapter_path>` to the LLM harness during startup

Existing behavior remains unchanged when no LoRA adapters are present.

---

### Linked issues

Fixes #306

### DIFF
--- a/pkg/cmd/dev/dev.go
+++ b/pkg/cmd/dev/dev.go
@@ -85,6 +85,28 @@ func runDev(ctx context.Context, options *DevStartOptions) error {
 		return err
 	}
 
+	// Detect and collect lora-adapter parts
+	var loraPaths []string
+	for _, part := range kitfile.Model.Parts {
+		if strings.EqualFold(part.Type, "lora-adapter") {
+			if part.Path == "" {
+				continue
+			}
+			partAbsPath, _, err := filesystem.VerifySubpath(options.contextDir, part.Path)
+			if err != nil {
+				output.Debugf("Skipping lora-adapter: %v", err)
+				continue
+			}
+			loraPath, err := findLoraAdapterFile(partAbsPath)
+			if err != nil {
+				output.Debugf("Skipping lora-adapter: %v", err)
+				continue
+			}
+			output.Infof("Found lora-adapter: %s", loraPath)
+			loraPaths = append(loraPaths, loraPath)
+		}
+	}
+
 	llmHarness := &harness.LLMHarness{}
 	llmHarness.Host = options.host
 	llmHarness.Port = options.port
@@ -93,7 +115,7 @@ func runDev(ctx context.Context, options *DevStartOptions) error {
 		return err
 	}
 
-	if err := llmHarness.Start(modelPath); err != nil {
+	if err := llmHarness.Start(modelPath, loraPaths); err != nil {
 		return err
 	}
 
@@ -168,6 +190,26 @@ func findModelFile(absPath string) (string, error) {
 	}
 	output.Debugf("Found model path in directory %s at %s", absPath, modelPath)
 	return modelPath, nil
+}
+
+// findLoraAdapterFile validates a lora adapter path.
+// The path must point to a regular file.
+func findLoraAdapterFile(absPath string) (string, error) {
+	stat, err := os.Lstat(absPath)
+	if err != nil {
+		return "", err
+	}
+
+	if !stat.Mode().IsRegular() {
+		return "", fmt.Errorf("lora adapter path must be a regular .gguf file: %s", absPath)
+	}
+
+	if !strings.HasSuffix(strings.ToLower(absPath), ".gguf") {
+		return "", fmt.Errorf("lora adapter file must be a .gguf file: %s", absPath)
+	}
+
+	output.Debugf("Found lora adapter path at %s", absPath)
+	return absPath, nil
 }
 
 // extractModelKitToCache extracts a ModelKit reference to a cache directory
--- a/pkg/lib/harness/llm-harness.go
+++ b/pkg/lib/harness/llm-harness.go
@@ -63,7 +63,7 @@ func (harness *LLMHarness) Init() error {
 	return nil
 }
 
-func (harness *LLMHarness) Start(modelPath string) (err error) {
+func (harness *LLMHarness) Start(modelPath string, loraPaths []string) (err error) {
 
 	harnessPath := constants.HarnessPath(harness.ConfigHome)
 	pidFile := filepath.Join(harnessPath, constants.HarnessProcessFile)
@@ -85,24 +85,31 @@ func (harness *LLMHarness) Start(modelPath string) (err error) {
 
 	uiHome := filepath.Join(harnessPath, "ui")
 	output.Debugf("model path is %s", modelPath)
+	for _, loraPath := range loraPaths {
+		output.Debugf("lora adapter path is %s", loraPath)
+	}
 	var cmd *exec.Cmd
+	args := []string{
+		"--server",
+		"--model", modelPath,
+		"--host", harness.Host,
+		"--port", fmt.Sprintf("%d", harness.Port),
+		"--path", uiHome,
+		"--gpu", "AUTO",
+		"--nobrowser",
+		"--unsecure",
+	}
+	for _, loraPath := range loraPaths {
+		args = append(args, "--lora", loraPath)
+	}
+
 	if runtime.GOOS == "windows" {
-		cmd = exec.Command(
-			"./llamafile.exe",
-			"--server",
-			"--model", modelPath,
-			"--host", harness.Host,
-			"--port", fmt.Sprintf("%d", harness.Port),
-			"--path", uiHome,
-			"--gpu", "AUTO",
-			"--nobrowser",
-			"--unsecure",
-		)
+		cmd = exec.Command("./llamafile.exe", args...)
 	} else {
-		cmd = exec.Command("sh", "-c",
-			fmt.Sprintf("./llamafile --server --model %s --host %s --port %d --path %s --gpu AUTO --nobrowser --unsecure",
-				modelPath, harness.Host, harness.Port, uiHome),
-		)
+		// Run through sh -c for APE compatibility, while passing all user-controlled
+		// values as positional args to avoid shell interpolation and injection.
+		shellArgs := append([]string{"-c", "exec ./llamafile \"$@\"", "llamafile"}, args...)
+		cmd = exec.Command("sh", shellArgs...)
 	}
 
 	cmd.Dir = harnessPath


pkg/lib/harness/llm-harness.go
```diff
@@ -97,11 +99,19 @@ func (harness *LLMHarness) Start(modelPath string) (err error) {
...
+		loraArgs := ""
+		for _, loraPath := range loraPaths {
+			loraArgs += fmt.Sprintf(" --lora %s", loraPath)
+		}
 		cmd = exec.Command("sh", "-c",
-			fmt.Sprintf("./llamafile --server --model %s --host %s --port %d --path %s --gpu AUTO --nobrowser --unsecure",
-				modelPath, harness.Host, harness.Port, uiHome),
+			fmt.Sprintf("./llamafile --server --model %s --host %s --port %d --path %s --gpu AUTO --nobrowser --unsecure%s",
+				modelPath, harness.Host, harness.Port, uiHome, loraArgs),
 		)
```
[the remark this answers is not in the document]
rishi-jat: Good catch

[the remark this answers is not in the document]
rishi-jat: fixed

amisevsk:
@rishi-jat there are some test failures on this PR

pkg/lib/harness/llm-harness.go
```diff
@@ -97,12 +99,26 @@ func (harness *LLMHarness) Start(modelPath string) (err error) {
...
+			"--port", fmt.Sprintf("%d", harness.Port),
+			"--path", uiHome,
+			"--gpu", "AUTO",
+			"--nobrowser",
+			"--unsecure",
+		}
+		for _, loraPath := range loraPaths {
+			args = append(args, "--lora", loraPath)
+		}
+		cmd = exec.Command("./llamafile", args...)
```
amisevsk: The arguments seem to be the same between the windows and non-windows case. Why not define them once and branch on the `exec.Command` step?

```go
	args := []string{
		"--server",
		"--model", modelPath,
		"--host", harness.Host,
		"--port", fmt.Sprintf("%d", harness.Port),
		"--path", uiHome,
		"--gpu", "AUTO",
		"--nobrowser",
		"--unsecure",
	}
	for _, loraPath := range loraPaths {
		args = append(args, "--lora", loraPath)
	}
	if runtime.GOOS == "windows" {
		cmd = exec.Command("./llamafile.exe", args...)
	} else {
		cmd = exec.Command("./llamafile", args...)
	}
```

pkg/cmd/dev/dev.go
```diff
@@ -170,6 +189,43 @@ func findModelFile(absPath string) (string, error) {
...
+			}
+		}
+		return nil
+	}); err != nil {
+		return "", fmt.Errorf("error searching for lora adapter file in %s: %w", absPath, err)
+	} else if loraPath == "" {
+		return "", fmt.Errorf("could not find lora adapter file in %s", absPath)
+	}
+	output.Debugf("Found lora adapter path in directory %s at %s", absPath, loraPath)
+	return loraPath, nil
```
amisevsk: Since only one LoRA is supported, I think it makes sense to avoid the walkDir step entirely and error out if the path specified by the modelpart is not a regular file.

pkg/cmd/dev/dev.go
```diff
@@ -170,6 +189,43 @@ func findModelFile(absPath string) (string, error) {
...
+			if loraPath == "" {
+				loraPath = path
+			} else {
+				return fmt.Errorf("multiple lora adapter files found: %s and %s", loraPath, path)
+			}
+		}
+		return nil
+	}); err != nil {
+		return "", fmt.Errorf("error searching for lora adapter file in %s: %w", absPath, err)
+	} else if loraPath == "" {
```
amisevsk: Unnecessary else:

```suggestion
	}
	if loraPath == "" {
```

rishi-jat: fixed.

pkg/lib/harness/llm-harness.go
rishi-jat: Good point.

pkg/cmd/dev/dev.go
rishi-jat: Updated.

rishi-jat:
@amisevsk I have made the suggested changes please have a look. Thanks!

amisevsk (approved):
Looks good to me, thanks @rishi-jat

pkg/cmd/dev/dev.go
```diff
@@ -84,6 +85,25 @@ func runDev(ctx context.Context, options *DevStartOptions) error {
 		return err
 	}
 
+	// Detect and collect lora-adapter parts
+	var loraPaths []string
+	for _, part := range kitfile.Model.Parts {
+		if strings.EqualFold(part.Type, "lora-adapter") {
+			partAbsPath, _, err := filesystem.VerifySubpath(options.contextDir, part.Path)
+			if err != nil {
+				output.Debugf("Failed to verify lora-adapter path %s: %v", part.Path, err)
```
gorkem: I think this should be Warning.  output.Debugf is only shown with --verbose. In normal operation the user sees "Found lora-adapter: ..." for successes but nothing for failures.

pkg/cmd/dev/dev.go
```diff
@@ -169,6 +189,26 @@ func findModelFile(absPath string) (string, error) {
 	return modelPath, nil
 }
 
+// findLoraAdapterFile validates a lora adapter path.
+// The path must point to a regular file.
+func findLoraAdapterFile(absPath string) (string, error) {
```
gorkem, line 197: `model.parts` can be either files or folders I think this function needs to support cases when it is a folder

gorkem (changes requested):
Thank you for the PR, this was something I was hoping to get to for a long time. PR needs to be rebased and there are a couple of issues that would be nice to address before merging.

pkg/cmd/dev/dev.go
rishi-jat: Good catch.

pkg/cmd/dev/dev.go
rishi-jat: +1

rishi-jat:
@gorkem @amisevsk  Made all the suggested changes have a look.

gorkem:
@rishi-jat the build fails. Please make sure that build passes cleanly

rishi-jat:
@gorkem CI pass. please have a look. Thanks!

pkg/lib/harness/llm-harness.go
```diff
@@ -139,17 +146,45 @@ func (harness *LLMHarness) Start(modelPath string) (err error) {
...
+	if err := os.MkdirAll(harnessPath, 0o755); err != nil {
+		return fmt.Errorf("error creating directory %s: %w", harnessPath, err)
+	}
+
+	if err := os.WriteFile(pidFile, []byte(mockHarnessPID), 0644); err != nil {
+		return fmt.Errorf("error writing PID to file %s: %w", pidFile, err)
+	}
+
+	return nil
+}
```
amisevsk: What is this function for? It does not make sense in `llm-harness.go`

amisevsk (changes requested):
I'm not sure why there's mock-related code in the llm-harness.go file

pkg/lib/harness/llm-harness.go
rishi-jat: Added for local testing, but it doesn’t belong in the harness.

rishi-jat: will remove it!

amisevsk (changes requested):
Tests are still failing

rishi-jat:
> Tests are still failing

CI is now green please have a look. Thanks!

gorkem (approved):

amisevsk (commented):
The current version is once again vulnerable to shell injection:

* We run with `--lora <LoRA-Path>` and `sh -c`, a malicious LoRA path can easily lead to shell injection
* Since we support LoRA adapters as directories (I still think it's not worth doing), it's additionally easy to obfuscate the presence of injection, as we can just store a directory path and put the malicious file in it to hide its presence from the user

E.g., if my ModelKit contains something like

```
/adapters/"lora.gguf &; echo "hi" > ~/hi.gguf"
```
and the Kitfile
```
model
  parts:
    - path: adapters/
      type: lora
```
you a) won't be able to notice an issue by looking at the Kitfile, and b) will automatically run arbitrary code when using `kit dev`.

It seems we're at an impasse here, as it seems `sh -c` is required for APE binaries used by llamafile, though.

amisevsk:
According to https://github.com/jart/cosmopolitan?tab=readme-ov-file#shells, the `sh -c` requirement was resolved in `zsh 5.9`. Assuming this is primarily a problem for macOS, I've checked an up-to-date mac and the default version of `zsh` there is 5.9 currently, though I don't know when this was updated.

Is the `sh -c` requirement still actual? zsh 5.9 was released in 2022.

pkg/lib/harness/llm-harness.go
```diff
@@ -97,11 +100,20 @@ func (harness *LLMHarness) Start(modelPath string) (err error) {
...
+		// Build command string for sh -c (required for APE binaries on Linux)
+		loraArgs := ""
+		for _, loraPath := range loraPaths {
+			loraArgs += fmt.Sprintf(" --lora %s", loraPath)
+		}
 		cmd = exec.Command("sh", "-c",
-			fmt.Sprintf("./llamafile --server --model %s --host %s --port %d --path %s --gpu AUTO --nobrowser --unsecure",
-				modelPath, harness.Host, harness.Port, uiHome),
+			fmt.Sprintf("./llamafile --server --model %s --host %s --port %d --path %s --gpu AUTO --nobrowser --unsecure%s",
+				modelPath, harness.Host, harness.Port, uiHome, loraArgs),
```
amisevsk: I understand reverting to using `sh -c` in 87f4adf, but why _also_ undo the change of using one args slice for both windows and unix? If we want the shell-injectable form here, we can still reuse the args slice rather than duplicating the arguments in this format:

```go
	args := []string{
		"--server",
		"--model", modelPath,
		"--host", harness.Host,
		"--port", fmt.Sprintf("%d", harness.Port),
		"--path", uiHome,
		"--gpu", "AUTO",
		"--nobrowser",
		"--unsecure",
	}

	for _, loraPath := range loraPaths {
		args = append(args, "--lora", loraPath)
	}

	if runtime.GOOS == "windows" {
		cmd = exec.Command("./llamafile.exe", args...)
	} else {
		// Note: running llamafile with sh -c is required as it is an APE binary (is this still true?)
		llamaCmd := "./llamafile " + strings.Join(args, " ")
		cmd = exec.Command("sh", "-c", llamaCmd)
	}
```

pkg/cmd/dev/dev.go
```diff
@@ -170,6 +192,50 @@ func findModelFile(absPath string) (string, error) {
...
+				}
+				found = path
+			}
+		}
+		if found == "" {
+			return "", fmt.Errorf("no .gguf lora adapter found in %s", absPath)
+		}
+		output.Debugf("Found lora adapter path in directory %s at %s", absPath, found)
+		return found, nil
+	}
```
amisevsk: I know me and @gorkem are giving conflicting review here, but I'm still of the opinion that if you have a model part with type `lora-adapter`, we should not support directories for this part. Given that a LoRA adapter is a specific thing, what's the use case for allowing it to be a directory?

amisevsk (changes requested):

rishi-jat:
Addressed the requested changes:

- removed shell-interpolated LoRA command construction in harness startup (still using sh -c for APE, now via safe positional args)
- unified startup argument construction across Windows/non-Windows
- restricted type: lora-adapter parts to regular .gguf files only (no directory resolution)
- rebased branch onto latest main

Could you please take another look?

PR #1099, opened 2026-02-27
fix: exclude OS-specific files from kit init Kitfile generation

### LINKED ISSUE #1081: `kit init` includes hidden files in Kitfile
**Describe the bug**
Running `kit init` will attempt to place _all_ files in the current directory inside the Kitfile. This includes hidden files, particularly OS-specific files like MacOS's `.DS_Store`. 

In some cases, this can lead to adding an entire opaque `code` section to the Kitfile solely to hold the hidden files.

Kit should

1. Ignore OS-specific files such as `.DS_Store` and similar in `kit init`; i.e. files not added by the user
1. Exclude OS-specific files even when a directory path might include them (e.g. a `code` layer with `path: .` that includes non-ignored iles)
1. (Maybe) Ignore packing hidden files by default, to be re-enabled via flag

**To Reproduce**
1. Run `kit init` in a directory containing hidden files
2. In all likelihood, notice a 'catch-all` `code` layer (`path: .`) that's added to include them, even when not necessary

**Version**
latest

**Additional context**
N/A

rishi-jat:
### Description
Running `kit init` previously included OS-generated metadata files (`.DS_Store`, `Thumbs.db`, etc.) in the generated Kitfile. This often resulted in unnecessary catch-all code layers (`path: .`) being added solely to include these hidden files.

## Changes

- Added `osSpecificFiles` list containing known OS-generated filenames (macOS and Windows)
- Added `shouldIgnoreFile()` helper function with case-insensitive matching
- Filter applied during directory scanning in `genDirListingFromPath()` to skip OS files before classification

## Files Ignored

| Filename | Platform |
|----------|----------|
| `.DS_Store` | macOS |
| `.AppleDouble` | macOS |
| `.LSOverride` | macOS |
| `.Spotlight-V100` | macOS |
| `.Trashes` | macOS |
| `Thumbs.db` | Windows |
| `desktop.ini` | Windows |
| `$RECYCLE.BIN` | Windows |
| `ehthumbs.db` | Windows |

## Testing

-  `kit init` with only OS files → no code layer generated
-  `kit init` with model + OS files → only model included, no catch-all
-  `kit init` with normal files → behavior unchanged
-  `kit pack` unaffected
-  All existing tests pass (`go test ./pkg/...`)

### Linked issues
Fixes #1081


gorkem:
The OS files that are now filtered from kit init will still be packed by `kit pack` if the Kitfile uses path: . The issue is explicitly asking to handle them. 

Have you considered ways to leverage the existing `.kitignore` mechanism ?

rishi-jat:
@gorkem Thanks for the feedback.

The scan logic now ignores hidden files and OS metadata files during Kitfile generation. This prevents the catch-all code layer (`path: .`) from being created due to files like `.DS_Store`.

The ignore helper is also exported as `generate.ShouldIgnoreFile()` so the same logic can be reused elsewhere if needed.

pkg/lib/kitfile/generate/filesystem.go
```diff
@@ -20,8 +20,37 @@ import (
...
+	"Thumbs.db",       // Windows
+	"desktop.ini",     // Windows
+	"$RECYCLE.BIN",    // Windows
+	"ehthumbs.db",     // Windows
+}
+
+// ShouldIgnoreFile returns true if the file should be ignored.
+// This helper is exported so other packages (for example packing logic)
+// can reuse the same OS-file filtering logic.
+func ShouldIgnoreFile(name string) bool {
```
amisevsk: The PR does not currently use this function in packing logic, so skipped files will still be included in ModelKits despite not being in a generated directory listing.

amisevsk (changes requested):

pkg/lib/kitfile/generate/filesystem.go
rishi-jat: fixed.

rishi-jat:
@amisevsk please have a look.

pkg/lib/repo/local/pull.go
```diff
@@ -254,6 +255,9 @@ func (l *localRepo) cleanupIngestDir() error {
 		if d.IsDir() {
 			return nil
 		}
+		if generate.ShouldIgnoreFile(d.Name()) {
+			return nil
+		}
```
amisevsk: It's strange to include the generate package in the pull package. It makes me think that `pkg/lib/kitfile/generate` is not the right place for this logic.

amisevsk: Additionally, why is this logic part of the cleanup process? Is there a bug we are not aware of?

amisevsk (changes requested):

pkg/lib/repo/local/pull.go
rishi-jat: I added this to keep OS file filtering consistent across flows, but looking at the actual flow, pull isn’t involved in this bug.

This was a mistake on my side - I’ll revert it and keep the fix scoped to generate and packing.

rishi-jat: fixed

rishi-jat:
@amisevsk

amisevsk (approved):

amisevsk:
@rishi-jat Sorry for the delay on reviewing this PR; could you rebase onto main so that we can merge?

rishi-jat:
> @rishi-jat Sorry for the delay on reviewing this PR; could you rebase onto main so that we can merge?

on it.

amisevsk (approved):

amisevsk:
Thanks @rishi-jat and sorry for the delays on this one!

PR #1100, opened 2026-03-01
Add pluggable Logger interface to output package

gorkem:
Embedding kit as a library means log output goes directly to the terminal, bypassing the host application's logging pipeline. There was no way to intercept or redirect it.

### What changed
- Added a Logger interface (pkg/output/interfaces.go) with a single Log(level, format, args) method
- Added SetLogger() in config.go to replace the default logger at runtime
- Introduced LogLevelSystem for messages that must always route to stderr (e.g. update notifications), independent of the configured log level
- Extracted formatAndWrite() helper to consolidate the formatting logic shared between defaultLogger and ProgressLogger

### What's not affected

ProgressLogger still writes directly to its underlying mpb writer — routing it through a custom logger would corrupt terminal progress bar output. Callers using a custom logger should also call SetProgressBars("none").

### Linked issues

https://github.com/kitops-ml/kitops/issues/905

### AI-Assisted Code

- [X] This PR contains AI-generated code that I have reviewed and tested
- [X] I take full responsibility for all code in this PR, regardless of how it was created

### DIFF
--- a/pkg/lib/hf/download.go
+++ b/pkg/lib/hf/download.go
@@ -112,7 +112,7 @@ func downloadFile(
 	}
 	defer func() {
 		if err := resp.Body.Close(); err != nil {
-			plog.Logf(output.LogLevelWarn, "Failed to close response body: %w", err)
+			plog.Logf(output.LogLevelWarn, "Failed to close response body: %s", err)
 		}
 	}()
 


pkg/output/logging.go
```diff
@@ -73,48 +94,24 @@ func SafeDebugf(s string, args ...any) {
...
-
-func loglnTo(output io.Writer, level LogLevel, s any) {
-	if logLevel.shouldPrint(level) {
-		str := fmt.Sprintln(s)
-		// Capitalize first letter in string for nicer output, in case it's not already capitalized
-		str = strings.ToUpper(str[:1]) + str[1:]
-		str = level.getPrefix() + str
-		fmt.Fprint(output, str)
-	}
+	currentLogger.Log(level, "%s\n", fmt.Sprint(s))
```
amisevsk: The signature for this log method is incorrect (from before this PR -- should be `...any`) but using `Sprint` (instead of `Sprintln`) here changes the semantics of how the `ln` methods work. If we want to simplify, we can just drop the `*ln` functions from the output package.


```suggestion
	currentLogger.Log(level, "%s", fmt.Sprintln(s))
```

gorkem:
I agree that this is a band-aid but it is pretty simple for what it is trying to accomplish, the other option was a big refactor that pretty much touched every .go file on the repo.

Issue #899, opened 2025-06-13
Support base digest pinning and OCI subject linkage for adapters


gorkem:
**Describe the problem you're trying to solve**
KitOps currently links adapters to base models via Kitfile metadata but does not record the exact base digest nor expose the relationship at the OCI level. This issue adds two enhancements to improve reliability and interoperability: capture and record the base’s immutable digest when packing adapters, and optionally set the OCI subject field in the adapter manifest to point at the base manifest digest.

**Describe the solution you'd like**
1. Digest pinning  - On kit pack for a Lora adapter, resolve the base reference (model.path) to its immutable OCI manifest digest
    - Record that digest in Config/Kitfile
1. OCI subject support
    - When generating the adapter’s OCI manifest, optionally set the subject field to reference the base manifest digest, for example:
      ```json
      "subject": {
        "mediaType": "application/vnd.oci.image.manifest.v1+json",
        "digest": "sha256:<base-manifest-digest>"
      }
      ```

arnab2001:
Willing to pick it up if its still needed

rishi-jat:
@gorkem  , Started working on this. Opened a PR - https://github.com/kitops-ml/kitops/pull/1126  for base digest pinning.

Will follow up with OCI subject linkage in a separate PR.

amisevsk:
Re-reading this issue, if a user cares about the base digest, can they not just specify the base model via its digest rather than its tag? I'm not sure this feature is adding anything useful to the project.

amisevsk:
I'm of the opinion that this issue should just be closed (cc: @gorkem)

gorkem:
I think this issue is touching the right problem but we should reconsider the solution.

The core need is real: when kit packs an adapter that references a base model by tag, we lose track of what that tag resolved to. Tags are mutable, digests are identity. We need to capture that resolution.

Docker build went through exactly this evolution. They started by encouraging users to pin digests manually in Dockerfiles (`FROM python@sha256:...`), which was bad UX if all you need is to record what it was built against (not avoid tag drift). Then BuildKit added `--provenance` which automatically records the resolved digest of every `FROM` image in a SLSA provenance attestation attached to the built image. The Dockerfile stays human-readable, the build record captures the cryptographic identity. We should learn from that rather than repeat the early mistakes.

I'd split this into two things instead of what's described here:

**1. Manifest annotation (always-on, zero config)**

[...]

PR #1126, opened 2026-03-18
feat(pack): add base digest pinning for adapters (best-effort)

rishi-jat:
### Description
Adds base digest pinning for adapters during `kit pack`.

Resolves the base model reference to its OCI manifest digest and stores it in `Model.BaseDigest`.  
Uses local-first resolution with remote fallback.  

This is best-effort only — pack never fails if resolution is not possible.  

Partially addresses #899

### DIFF
--- a/pkg/cmd/pack/pack.go
+++ b/pkg/cmd/pack/pack.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/kitops-ml/kitops/pkg/artifact"
 	cmdoptions "github.com/kitops-ml/kitops/pkg/cmd/options"
@@ -128,6 +129,10 @@ func pack(ctx context.Context, opts *packOptions, kitfile *artifact.KitFile, loc
 }
 
 func resolveBaseDigest(ctx context.Context, configHome, baseRef string) (string, error) {
+	if strings.HasPrefix(baseRef, "sha256:") {
+		return baseRef, nil
+	}
+
 	ref, _, err := artifact.ParseReference(baseRef)
 	if err != nil {
 		return "", fmt.Errorf("failed to parse base reference: %w", err)
@@ -142,7 +147,7 @@ func resolveBaseDigest(ctx context.Context, configHome, baseRef string) (string,
 	if err == nil {
 		return desc.Digest.String(), nil
 	}
-	// Expected cases: fallback to remote
+	// Expected cases: no base digest resolved locally (skip pinning, no remote fallback)
 	if errors.Is(err, util.ErrNoKitfile) || errors.Is(err, util.ErrNotAModelKit) {
 		return "", nil
 	}


pkg/cmd/pack/pack.go
```diff
@@ -112,6 +122,37 @@ func pack(ctx context.Context, opts *packOptions, kitfile *artifact.KitFile, loc
...
+
+	repository, err := remote.NewRepository(ctx, ref.Registry, ref.Repository, cmdoptions.DefaultNetworkOptions(configHome))
+	if err != nil {
+		return "", nil
+	}
+
+	desc, _, _, err = util.ResolveManifestAndConfig(ctx, repository, ref.Reference)
+	if err != nil {
+		return "", nil
+	}
```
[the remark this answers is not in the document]
rishi-jat: fixed

rishi-jat:
> See comments from copilot

Done, have a look.

pkg/cmd/pack/pack.go
```diff
@@ -86,6 +90,13 @@ func pack(ctx context.Context, opts *packOptions, kitfile *artifact.KitFile, loc
 			return nil, fmt.Errorf("Failed to resolve referenced modelkit %s: %w", kitfile.Model.Path, err)
 		}
 		extraLayerPaths = util.LayerPathsFromKitfile(parentKitfile)
+
+		baseDigest, err := resolveBaseDigest(ctx, opts.configHome, kitfile.Model.Path)
+		if err != nil {
+			output.Debugf("Skipping base digest resolution: %v", err)
+		} else if baseDigest != "" {
+			kitfile.Model.BaseDigest = baseDigest
```
gorkem, line 103: If a user manually pins baseDigest: sha256:abc123 in their Kitfile for reproducibility, every subsequent kit pack silently replaces it with the freshly resolved digest. At minimum this should be documented as deliberate behavior and we should discourage users from setting the baseDigest themselves or guarded with a check and a log like ("overriding user-specified baseDigest with freshly
resolved value").

pkg/cmd/pack/pack.go
```diff
@@ -86,6 +90,13 @@ func pack(ctx context.Context, opts *packOptions, kitfile *artifact.KitFile, loc
...
+		}
```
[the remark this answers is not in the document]
gorkem: Valid concern

pkg/cmd/pack/pack.go
```diff
@@ -112,6 +123,47 @@ func pack(ctx context.Context, opts *packOptions, kitfile *artifact.KitFile, loc
...
+		return "", nil
+	}
+	if errors.Is(err, errdef.ErrNotFound) || errors.Is(err, errdef.ErrMissingReference) {
+		// Reference not found locally, try remote
+	} else {
+		// Unexpected local error (IO issues, corruption, etc.)
+		return "", fmt.Errorf("failed to resolve base digest locally: %w", err)
+	}
+
+	repository, err := remote.NewRepository(ctx, ref.Registry, ref.Repository, cmdoptions.DefaultNetworkOptions(configHome))
```
gorkem, line 161: packOptions has no NetworkOptions field, so DefaultNetworkOptions is used with hardcoded defaults. Users behind registries requiring --plain-http, self-signed certs, or proxies will silently fail to pin the base digest with no actionable feedback. 

The scenario:

 1. User's Kitfile has baseDigest: sha256:aaa (set manually or by a prior run)
 2. Base model is updated → real digest is now sha256:bbb
 3. Pack is run but remote resolution fails (network unavailable, TLS mismatch, etc.)
 4. err != nil → debug-logged, BaseDigest untouched
 5. Packed artifact is written with baseDigest: sha256:aaa — pointing to the wrong base

So the PR's best-effort guarantee protects pack from failing, but it doesn't protect the packed artifact from being silently incorrect. The artifact claims to be derived from sha256:aaa but may actually be built on the updated base.

rishi-jat: Makes sense

rishi-jat:
@gorkem Addressed all comments.

pkg/cmd/pack/pack.go
```diff
@@ -86,6 +91,17 @@ func pack(ctx context.Context, opts *packOptions, kitfile *artifact.KitFile, loc
...
+
+		baseDigest, err := resolveBaseDigest(ctx, opts.configHome, kitfile.Model.Path)
+		if err != nil {
+			output.Logf(output.LogLevelWarn, "failed to resolve base digest, keeping existing value: %v", err)
+		} else if baseDigest != "" {
+			if kitfile.Model.BaseDigest != "" && kitfile.Model.BaseDigest != baseDigest {
+				output.Logf(output.LogLevelWarn, "overriding user-specified baseDigest %s with resolved %s",
+					kitfile.Model.BaseDigest, baseDigest)
+			}
+			kitfile.Model.BaseDigest = baseDigest
```
amisevsk, line 103: What's the point of being able to specify a base digest if it just gets overridden, pushing the digest mismatch into a single warning log line at build time.

Imagine running this build in a CI -- will you notice that the digest has changed? If not, what was the point in including it?

pkg/cmd/pack/pack.go
```diff
@@ -112,6 +128,51 @@ func pack(ctx context.Context, opts *packOptions, kitfile *artifact.KitFile, loc
...
+	// Expected cases: no base digest resolved locally (skip pinning, no remote fallback)
+	if errors.Is(err, util.ErrNoKitfile) || errors.Is(err, util.ErrNotAModelKit) {
+		return "", nil
+	}
+	if errors.Is(err, errdef.ErrNotFound) || errors.Is(err, errdef.ErrMissingReference) {
+		// Reference not found locally, try remote
+	} else {
+		// Unexpected local error (IO issues, corruption, etc.)
+		return "", fmt.Errorf("failed to resolve base digest locally: %w", err)
+	}
```
amisevsk, line 159: The comment here contradicts the code that follows.

rishi-jat:
@amisevsk sorry for late - would look into the review asap and do the suggested changes.Thanks!

PR #1127, opened 2026-03-19
Feature/issue 419 config cmd

omarbassem128:
### Description
This PR introduces the new `config` command and natively integrates a `config.yaml` file into the root CLI options, allowing users to persist global settings without needing to pass flags on every run. 

**Added Subcommands:**
* **`config list`**: Outputs the entire active configuration state in a formatted JSON structure. Evaluates the complete precedence hierarchy to show the exact settings the CLI is using.
* **`config set <key> <value>`**: Writes a new configuration value directly to the `config.yaml` file to persist global settings across terminal sessions.
* **`config get <key>`**: Retrieves and prints the value of a single, specific configuration setting.
* **`config reset`**: Deletes the `config.yaml` file, completely clearing any persisted user settings and instantly reverting the CLI back to its default values.

**Key Implementations:**
* **Global State Management:** Integrated the config file reader into the `PersistentPreRunE` hook so that `log-level`, `progress`, and `verbosity` are properly initialized for all subcommands.
* **Strict Precedence Hierarchy:** Built a robust override system to respect standard terminal behavior:
  * **Storage Path (`configHome`):** CLI Flag (`--config`) > Environment Variable (`$KITOPS_HOME`) > Config File > Defaults.
  * **Settings (`log-level`, etc.):** CLI Flags > Config File > Defaults.
* **Load Sequence Architecture:** Refactored the boot sequence to ensure the CLI dynamically resolves the correct target storage directory *before* attempting to load the YAML file.
* **Graceful Degradation & Zero-Value Handling:** Handled empty configuration files safely by falling back to Cobra's default options.
  * Replaced fatal crashes with non-fatal warnings when the config file contains invalid strings, allowing the application to safely apply default values and continue running.

### Linked issues
- Closes: kitops-ml#419

### AI-Assisted Code

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -72,6 +72,8 @@ func RunCommand() *cobra.Command {
 		Short: shortDesc,
 		Long:  longDesc,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			output.SetOut(cmd.OutOrStdout())
+			output.SetErr(cmd.ErrOrStderr())
 
 			configHome, err := getConfigHomePath(opts)
 			if err != nil {
@@ -86,9 +88,6 @@ func RunCommand() *cobra.Command {
 			if loadErr != nil && !errors.Is(loadErr, os.ErrNotExist) {
 				return loadErr
 			}
-
-			output.SetOut(cmd.OutOrStdout())
-			output.SetErr(cmd.ErrOrStderr())
 
 			if cmd.Flags().Changed("log-level") || configStruct.LogLevel == "" {
 				if err := output.SetLogLevelFromString(opts.loglevel); err != nil {


omarbassem128:
@amisevsk Appreciate the feedback.

pkg/cmd/config/config.go
```diff
@@ -0,0 +1,144 @@
...
+
+func PathHelper() (string, error) {
+	path, err := constants.DefaultConfigPath()
+	if err != nil {
+		return "", fmt.Errorf("failed to get default path: %w", err)
+	}
+	configYamlPath := constants.ConfigYamlPath(path)
+
+	return configYamlPath, nil
+}
```
amisevsk: How do you handle cases where a non-default config path is being used? I can override the default config path by setting either `$KITOPS_HOME` or using the `--config` flag.

cmd/root.go
```diff
@@ -71,35 +72,51 @@ func RunCommand() *cobra.Command {
...
 			}
+
+			configYamlPath := constants.ConfigYamlPath(configHome)
+			configStruct, loadErr := config.LoadConfigFileHelper(configYamlPath)
+			if loadErr != nil && !errors.Is(loadErr, os.ErrNotExist) {
+				return loadErr
+			}
+
+			output.SetOut(cmd.OutOrStdout())
+			output.SetErr(cmd.ErrOrStderr())
```
amisevsk: You have already done some logging; these should be set before we do any output

pkg/cmd/config/config.go
omarbassem128: Yes this should be resolved by using the context set in the root.go file. Currently working on it.

omarbassem128: I found some bugs that I am trying to fix while passing the context. Which is why it is taking some time.

omarbassem128: I solved it and other bugs in this new [commit](https://github.com/kitops-ml/kitops/pull/1127/changes/5d6d204c93d884718d81343f13888fa1ffaccd4d)

cmd/root.go
omarbassem128: Moved them to the top

pkg/cmd/config/config.go
omarbassem128: Having `configHome` in the config.yaml led to a very confusing chicken-and-egg situation which is why I removed it completely. What do you think ? Now if the user wants a custom path to their config file, they should use the `--config` flag with every command, whether it is `set`, `list`, `get` or `reset`.

omarbassem128: Quick correction to my reply above after doing a final review of the architecture. I completely forgot that we already have the `$KITOPS_HOME` environment variable wired up perfectly in root.go. Specifically in `getConfigHomePath()`. So, the workflow is actually exactly what we want. Users can use the `--config` flag for one-off ephemeral commands, and set the `$KITOPS_HOME` environment variable for persistent custom paths. I was originally trying to think of a way to avoid forcing users to type `--config` every time (since that defeats the purpose of the issue), but the environment variable already handles this natively. I think relying on that existing env variable is the right way to go.

omarbassem128:
@amisevsk Would you kindly check the recent changes ?

Issue #784, opened 2025-02-27
Add option to generate an OCI container image


notniknot:
More and more tools and features arise that support models as OCI compliant artifacts.
- Kubernetes will allow mounting container images as volumes from Version 1.31 onwards which is more elegant than InitContainers
- KServe supports "ModelCars" which is similar. An inference service, based on a model in a container image, will be deployed.

Unfortunately, ModelKits cannot be natively used with those tools/features. Kubernetes throws `Failed to pull image ...: parsing image configuration: unsupported image-specific operation on artifact with type "application/vnd.kitops.modelkit.config.v1+json"` and KServe ModelCar-Images need a shell to function.
We are using Red Hat OpenShift AI which introduces KServe and ModelCars as well. Therefore, we would like to use KitOps as a tool to build those "ModelCars". Otherwise, building an OCI image out of an OCI artifact (ModelKit) by introducing another tool or process seems to me a little bit too complex.

An option to package the ModelKit into an OCI Image, which is based on a micro image that the user can choose could be a great option to make KitOps more interoperable.

gorkem:
@notniknot Have you checked https://github.com/jozu-ai/kitops/tree/main/build/dockerfiles/KServe to see if this would provide the integration that you need with KServe. 

We are also working under the CNCF umbrella with other organizations to create a standard similar to ModelKits that will allow the whole CNCF to get behind.

amisevsk:
There is also the [`kitops-init`](https://github.com/jozu-ai/kitops/pkgs/container/kitops-init) container (documentation [here](https://github.com/jozu-ai/kitops/tree/main/build/dockerfiles/init))

On the other hand, it would not be terribly difficult to convert a ModelKit into a OCI container, so this is something we could pursue.

notniknot:
Thanks @gorkem and @amisevsk for the information! I did not know about ClusterStorageContainers.

But unfortunately, as far as I can tell, Red Hat OpenShift AI only supports paths to either S3 storage or an OCI image (see this blog post as example https://developers.redhat.com/articles/2025/01/30/build-and-deploy-modelcar-container-openshift-ai#deploy_a_model_server_with_a_modelcar_container).
If it wouldn't be too difficult, being able to convert a ModelKit to an OCI Container with customizable base image would still be really amazing.

gorkem:
[...]

PR #1128, opened 2026-03-19
Add local OCI image builder 

Arpit529Srivastava:
### Description
so this pr adds a new go package `pkg/lib/filesystem/ociimage` that i wrote to build a valid oci container image from a local directory and write it out as an oci layout on disk. the image it produces uses `application/vnd.oci.image.config.v1+json` for the oci image config and places all files under `/models`, which is really the core foundation i needed to get kserve modelcar and kubernetes image volume compatibility working correctly.

### Linked issues

ref #784 
### AI-Assisted Code

### DIFF
--- a/pkg/lib/ociimage/image.go
+++ b/pkg/lib/ociimage/image.go
@@ -1,0 +1,335 @@
+// Copyright 2024 The KitOps Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


Arpit529Srivastava:
> I think building and managing full OCI images is out of scope for this project. There is enough complexity in handling these flows properly that I do not think it suitable to include them in KitOps.

okay, closing the PR as @amisevsk pointed out, this is out of scope. will open this if there is any future discussion on this implementation.

PR #1130, opened 2026-03-21
fix: support different digest algorithms for pulling and storage

### LINKED ISSUE #414: Add tests to support different Digest algorithms
Kit defaults to SHA256 for digests. Develop a set of tests and fix when necessary to support multiple digest algorithms.

Arpit529Srivastava:
### Description
addressing the hardcoded `sha256` digest directory paths introduced during the implementation of resumable downloads #408. while oci commonly defaults to sha256, the specification allows other algorithms (like sha512). this change updates the implementation to dynamically extract the digest algorithm from `ocispec.Descriptor` across the local repository stack, ensuring full oci compliance and compatibility with any valid hashing algorithm.

### Linked issues
fixes #414

### AI-Assisted Code

### DIFF
--- a/pkg/lib/repo/local/digest_test.go
+++ b/pkg/lib/repo/local/digest_test.go
@@ -1,0 +1,135 @@
+// Copyright 2024 The KitOps Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


Arpit529Srivastava:
/cc @gorkem @amisevsk PTAL,
thanks :)

PR #1133, opened 2026-03-23
feat(pack): add optional OCI subject linkage for adapters

rishi-jat:
Adds optional OCI subject linkage during pack.

- Resolves base manifest descriptor (local-first, remote fallback)
- Sets manifest.Subject when available
- Best-effort: does not fail pack on resolution errors

Part of #899

### DIFF
--- a/pkg/cmd/pack/pack.go
+++ b/pkg/cmd/pack/pack.go
@@ -18,21 +18,27 @@ package pack
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/kitops-ml/kitops/pkg/artifact"
+	cmdoptions "github.com/kitops-ml/kitops/pkg/cmd/options"
 	"github.com/kitops-ml/kitops/pkg/lib/constants"
 	"github.com/kitops-ml/kitops/pkg/lib/constants/mediatype"
 	"github.com/kitops-ml/kitops/pkg/lib/filesystem"
 	"github.com/kitops-ml/kitops/pkg/lib/filesystem/ignore"
 	kfutils "github.com/kitops-ml/kitops/pkg/lib/kitfile"
 	"github.com/kitops-ml/kitops/pkg/lib/repo/local"
+	"github.com/kitops-ml/kitops/pkg/lib/repo/remote"
 	"github.com/kitops-ml/kitops/pkg/lib/repo/util"
 	"github.com/kitops-ml/kitops/pkg/output"
 
+	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2/errdef"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/cmd/pack/pack.go
```diff
@@ -105,13 +119,61 @@ func pack(ctx context.Context, opts *packOptions, kitfile *artifact.KitFile, loc
...
 	}
 	return manifestDesc, nil
 }
 
+func resolveBaseManifestDescriptor(ctx context.Context, configHome, baseRef string) (*ocispec.Descriptor, error) {
+	if strings.HasPrefix(baseRef, "sha256:") {
+		parsedDigest, err := digest.Parse(baseRef)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse digest-only base reference: %w", err)
+		}
```
amisevsk, line 135: In what cases is `baseRef` a bare string? I don't think that's a valid option for `model.path` -- I don't know that it would even satisfy `artifact.IsModelKitReference(kitfile.Model.Path)`.

Copilot's comment is also valid on this block, but it's secondary to it being what appears to be dead code.

pkg/cmd/pack/pack.go
```diff
@@ -105,13 +119,61 @@ func pack(ctx context.Context, opts *packOptions, kitfile *artifact.KitFile, loc
...
+
+	localRepo, err := local.NewLocalRepo(constants.StoragePath(configHome), ref)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create local repository: %w", err)
+	}
+
+	desc, _, _, err := util.ResolveManifestAndConfig(ctx, localRepo, ref.Reference)
+	if err == nil {
+		return &desc, nil
+	}
```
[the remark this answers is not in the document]
amisevsk: Copilot is correct, if we're discarding the config we shouldn't be fetching it in the first place.

pkg/cmd/pack/pack.go
```diff
@@ -105,13 +119,61 @@ func pack(ctx context.Context, opts *packOptions, kitfile *artifact.KitFile, loc
...
+	if errors.Is(err, util.ErrNoKitfile) || errors.Is(err, util.ErrNotAModelKit) {
+		return nil, nil
+	}
+	if !errors.Is(err, errdef.ErrNotFound) && !errors.Is(err, errdef.ErrMissingReference) {
+		return nil, fmt.Errorf("failed to resolve base descriptor locally: %w", err)
+	}
+
+	repository, err := remote.NewRepository(ctx, ref.Registry, ref.Repository, cmdoptions.DefaultNetworkOptions(configHome))
```
amisevsk, line 160: Using the `defaultNetworkOptions` here is rather fragile; if any non-default options are required, this will fail. On the other hand, traditionally in a pack context we don't have network options available.

This is probably fine for now, though likely should be improved at some point in the future.

amisevsk (changes requested):

PR #1139, opened 2026-03-25
Fix #1138: Set org.opencontainers.image.created annotation on ModelKit manifests

### LINKED ISSUE #1138: Set org.opencontainers.image.created annotation on ModelKit manifests
When kit packs a ModelKit, it sets a CreatedAt timestamp in the OCI config blob (ModelDescriptor) but does not set the standard `org.opencontainers.image.created` annotation on the manifest itself.

Tools and registries that inspect OCI manifests for creation time (e.g. registry UIs, garbage collection policies, lifecycle management tooling) rely on the [pre-defined OCI annotation](https://github.com/opencontainers/image-spec/blob/main/annotations.md) `org.opencontainers.image.created`and won't find it on ModelKits today.

Set `org.opencontainers.image.created` in the manifest annotations during pack, alongside the existing `ml.kitops.modelkit.cli-version` annotation in pkg/lib/filesystem/local-storage.go.

The value should be the current time in RFC 3339 format, per the OCI image spec.

itniuma2026:
## Summary

Add the standard OCI annotation `org.opencontainers.image.created` with the current time in RFC 3339 format to ModelKit manifests during the pack operation, alongside the existing `ml.kitops.modelkit.cli-version` annotation.

## Approach

In the function within `pkg/lib/filesystem/local-storage.go` that constructs the manifest and sets annotations (where `ml.kitops.modelkit.cli-version` is already set), add a new annotation entry with key `org.opencontainers.image.created` and value set to `time.Now().UTC().Format(time.RFC3339)`. This requires importing the `time` package if not already imported. The change should be placed right next to the existing CLI version annotation assignment to keep annotation logic grouped together.

## Files Changed

- `pkg/lib/filesystem/local-storage.go`

## Related Issue

Fixes #1138

## Testing

No tests were added with this change. Happy to add them if needed.

### DIFF
--- a/pkg/lib/filesystem/local-storage.go
+++ b/pkg/lib/filesystem/local-storage.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/kitops-ml/kitops/pkg/artifact"
 	"github.com/kitops-ml/kitops/pkg/lib/constants"
@@ -63,6 +64,11 @@ func SaveModel(ctx context.Context, localRepo local.LocalRepo, kitfile *artifact
 	if err != nil {
 		return nil, fmt.Errorf("error creating manifest: %w", err)
 	}
+
+	if manifest.Annotations == nil {
+		manifest.Annotations = map[string]string{}
+	}
+	manifest.Annotations[ocispec.AnnotationCreated] = time.Now().UTC().Format(time.RFC3339)
 
 	// If not storing a Kitfile, save the Kitfile to an annotation since it will be lost otherwise
 	if opts.ModelFormat == mediatype.ModelPackFormat {
--- a/pkg/lib/filesystem/local-storage.go
+++ b/pkg/lib/filesystem/local-storage.go
@@ -65,11 +65,6 @@ func SaveModel(ctx context.Context, localRepo local.LocalRepo, kitfile *artifact
 		return nil, fmt.Errorf("error creating manifest: %w", err)
 	}
 
-	if manifest.Annotations == nil {
-		manifest.Annotations = map[string]string{}
-	}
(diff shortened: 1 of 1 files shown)


gorkem:

gorkem:
@itniuma2026 can you fix the DCO failure

itniuma2026:
@gorkem Done! I've amended the commit with the required `Signed-off-by` line. The DCO check should pass now. Thanks for the reminder!

pkg/lib/filesystem/local-storage.go
```diff
@@ -64,6 +65,11 @@ func SaveModel(ctx context.Context, localRepo local.LocalRepo, kitfile *artifact
 		return nil, fmt.Errorf("error creating manifest: %w", err)
 	}
 
+	if manifest.Annotations == nil {
+		manifest.Annotations = map[string]string{}
+	}
+	manifest.Annotations[ocispec.AnnotationCreated] = time.Now().UTC().Format(time.RFC3339)
+
```
amisevsk: `createManifest` already sets up _some_ annotations (the CLI version) -- it should probably also apply this annotation as part of that process. The reason we set the additional annotation below is because it's optionally included and ModelPack-specific behaviour.

`createManifest` should be responsible for creating the base `ModelKit` manifest in as complete a way as possible.

amisevsk (changes requested):

pkg/lib/filesystem/local-storage.go
itniuma2026: Good point! Moved the `org.opencontainers.image.created` annotation into `createManifest()` alongside the CLI version annotation, and removed the redundant nil-check from `SaveModel()`. Thanks for the review.

amisevsk (approved):

amisevsk:
Thanks @itniuma2026!

PR #1141, opened 2026-03-25
fix(git): check destination emptiness without full readdir

AdeshDeshmukh:
Avoid reading an entire directory just to determine if clone destination is empty. Use ReadDir(1) to detect emptiness efficiently while preserving error behavior for non-empty and inaccessible paths.

### Description

This PR improves clone destination validation in `pkg/lib/external/git/clone.go`.

Previously, `checkDestination` used `os.ReadDir(path)` to determine whether a directory is empty, which reads the entire directory contents even though only an empty/non-empty check is needed.

This change replaces that with a one-entry probe:
- `os.Open(path)`
- `ReadDir(1)`

Behavior is preserved:
- `err == nil` => directory is non-empty
- `err == io.EOF` => directory is empty
- other errors => return wrapped inspection error

Validation:
- `go test ./pkg/lib/external/git` (pass)

### Linked issues

N/A

### DIFF
--- a/pkg/lib/external/git/clone.go
+++ b/pkg/lib/external/git/clone.go
@@ -19,6 +19,7 @@ package git
 import (
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"os"
 	"os/exec"
@@ -113,13 +114,19 @@ func checkDestination(path string) error {
 	if !stat.IsDir() {
 		return fmt.Errorf("path %s exists and is not a directory", path)
 	}
-	// TODO: probably don't need to read the _whole_ directory
-	contents, err := os.ReadDir(path)
+	// Only check whether at least one entry exists; avoid reading full directory.
+	dir, err := os.Open(path)
 	if err != nil {
 		return fmt.Errorf("failed to inspect directory %s: %w", path, err)
 	}
-	if len(contents) > 0 {
+	defer dir.Close()
+
+	_, err = dir.ReadDir(1)
+	if err == nil {
 		return fmt.Errorf("cannot clone to a non-empty directory")
+	}
+	if !errors.Is(err, io.EOF) {
+		return fmt.Errorf("failed to inspect directory %s: %w", path, err)
 	}
 	return nil
 }


gorkem:

gorkem:
@AdeshDeshmukh please fix failing DCO check before we can review this.

pkg/lib/constants/consts.go
```diff
@@ -49,15 +49,21 @@ const (
...
+	LegacyCliVersionAnnotation  = "ml.kitops.modelkit.cli-version"
+	LegacyKitfileJsonAnnotation = "ml.kitops.modelkit.kitfile"
 
 	// LayerSubtypeAnnotation stores additional type information for a given OCI manifest
 	// layer within its annotations (e.g. storing prompts within code-type layers)
-	LayerSubtypeAnnotation = "ml.kitops.modelkit.layer-subtype"
-	LayerSubtypePrompt     = "prompt"
+	LayerSubtypeAnnotation       = "org.kitops.modelkit.layer-subtype"
+	LegacyLayerSubtypeAnnotation = "ml.kitops.modelkit.layer-subtype"
+	LayerSubtypePrompt           = "prompt"
```
amisevsk: This change, while generally something we probably should do, is unrelated to the purpose of this PR and should be broken out into a separate PR. 

Additionally, it adds a decent amount of complexity for handling annotations; I think we're fine sticking with the `ml` prefix for now.

amisevsk (changes requested):
Please reduce this changeset to only address the issue mentioned in the PR title; changing annotations used by Kit and adding legacy options should be done in a separate PR (personally I don't think this is very high priority at the moment).

Also, there are merge conflicts that need to be addressed.

AdeshDeshmukh:
@amisevsk Thank you for your thorough review and valuable suggestions. I've removed the annotation migration changes as you recommended — the PR now focuses solely on the clone.go directory emptiness check. I've also resolved the merge conflicts and ensured DCO compliance. Would appreciate a re-review when you a bandwidth. PTAL

amisevsk (approved):

PR #1142, opened 2026-03-25
fix: generate Kitfile for ModelPack if missing from manifest

AdeshDeshmukh:
Use fallback generation for missing Kitfile in ModelPacks.

### Description

This PR introduces a fallback mechanism to generate a Kitfile in-memory when a ModelPack manifest does not contain one in its annotations. 

Previously, if the `org.cncf.model.kitfile` annotation was missing, `GetKitfileForManifest` would return `ErrNoKitfile`. This change adds `GenerateKitfileForModelPack` to reconstruct the Kitfile from layer metadata (filepaths and media types), allowing tools to inspect and unpack these artifacts seamlessy.

Validation:
- Added new unit tests in `pkg/lib/repo/util/modelpack_test.go`
- Verified existing logic in `pkg/lib/filesystem/unpack/core.go` now uses the shared utility.

### Linked issues

Fixes missing Kitfile handling for ModelPacks.

### DIFF
--- a/pkg/lib/filesystem/unpack/core.go
+++ b/pkg/lib/filesystem/unpack/core.go
@@ -37,7 +37,6 @@ import (
 	"github.com/kitops-ml/kitops/pkg/lib/repo/util"
 	"github.com/kitops-ml/kitops/pkg/output"
 
-	modelspecv1 "github.com/modelpack/model-spec/specs-go/v1"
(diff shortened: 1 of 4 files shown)


gorkem:

amisevsk:
Given the lack of attention on this PR, and as equivalent functionality was added in https://github.com/kitops-ml/kitops/pull/1202, I'm closing this PR.

PR #1164, opened 2026-04-16
Add ModelPack created annotation to manifests

### LINKED ISSUE #1138: Set org.opencontainers.image.created annotation on ModelKit manifests

gnanirahulnutakki:
Closes #1138

This adds `org.opencontainers.image.created` to ModelPack manifests only, and it reuses the same `CreatedAt` value already emitted into the ModelPack config instead of minting a second timestamp at manifest creation time. That keeps the manifest aligned with the existing ModelPack metadata while leaving KitFormat behavior unchanged.

Validation:
- `go test ./pkg/lib/filesystem -run TestCreateManifest`
- `go build ./...`
- `go generate ./...` (completed locally; the harness UI generator emitted `pnpm` warnings in this environment but produced no tracked source changes)

### DIFF
--- a/pkg/lib/constants/consts.go
+++ b/pkg/lib/constants/consts.go
@@ -51,8 +51,9 @@ const (
 
 	// Kitops-specific annotations for modelkit artifacts
 	// TODO: update these to use the newer kitops.org domain
-	CliVersionAnnotation  = "ml.kitops.modelkit.cli-version"
-	KitfileJsonAnnotation = "ml.kitops.modelkit.kitfile"
+	CliVersionAnnotation      = "ml.kitops.modelkit.cli-version"
+	KitfileJsonAnnotation     = "ml.kitops.modelkit.kitfile"
+	OciImageCreatedAnnotation = "org.opencontainers.image.created"
 
 	// LayerSubtypeAnnotation stores additional type information for a given OCI manifest
 	// layer within its annotations (e.g. storing prompts within code-type layers)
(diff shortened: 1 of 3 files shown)


gnanirahulnutakki:
Pushed a follow-up on April 30, 2026 to fix the failing `Check sources` job on this branch.

The failure was mechanical: the new Go test file was missing the required Apache license header. Current local recheck on the updated branch:

```bash
/opt/homebrew/bin/go test ./pkg/lib/filesystem
$HOME/go/bin/addlicense --check -s -l apache -c "The KitOps Authors." pkg/lib/filesystem/local-storage_test.go
```

Both pass locally.

The new GitHub workflow runs on the fork are currently showing `action_required` with no jobs started yet on my side, so it looks like they may need maintainer approval before the updated PR jobs execute.

PR #1175, opened 2026-05-05
Add --attestation-output flag to kit import for SLSA Provenance v1

gorkem:
## Summary

Adds an opt-in `--attestation-output <path>` flag on `kit import` that emits a SLSA Provenance v1 predicate for the imported ModelKit. Output is byte-compatible with `cosign attest --predicate <path> --type slsaprovenance1`. Default behavior is unchanged.

The predicate pins:
- `source.uri` — `hf://[datasets/]<repo>` or `git+https://host/repo` (userinfo scrubbed from logs and predicate)
- `source.digest.gitCommit` — HF: `PinCommit` via `/api/.../revision/{ref}` before listing/downloading; git: `ResolveHead` on the local clone
- `kitfile.digest.sha256` — the OCI config blob hash, byte-identical to what cosign attests
- manifest digest — logged to stderr; cosign derives the in-toto Statement subject from the OCI ref

`resolvedDependencies` is intentionally `[]` — the source commit and manifest digests already form a complete integrity chain, so per-file enumeration would just restate what's already pinned.

## Other changes

- **Atomic data sink** (`pkg/output/sink.go`): `OpenDataSink` writes file targets through a temp file in the same directory, atomically renamed on `Commit`. A failing import leaves the user's existing attestation file untouched. For `-`, the info writer reroutes to stderr so logs don't corrupt the JSON. `kit init` also uses this helper.
- **Commit pinning before listing** (HF): `PinCommit` runs first so `ListFiles` and `DownloadFiles` both bind to one immutable SHA — closes the TOCTOU window where a branch could move mid-import.
- **Path-traversal guard**: `DownloadFiles` runs `filesystem.VerifySubpath` on every entry before any network/filesystem work, rejecting `..`-bearing paths from a malicious or compromised HF response.
- **Layer-aware download filter**: `kitfilePathFilter` mirrors pack's ignore-aware layer routing so HF only downloads files that will end up in the ModelKit. Handles catchall layers, `.kitignore` patterns (which always cover Kitfile names and `.kitignore` itself), Windows-style separators in cleaned layer paths, and the prefix-boundary case.
- **URL scrubbing**: `git.SanitizeURL` strips userinfo from clone URLs in both the `Cloning repository …` log line and the predicate `SourceURI`.
- **Validation**: `(*ProvenanceData).validate()` refuses to emit a predicate with empty source/kitfile/manifest digests, so an unverifiable attestation can't slip through silently.

### DIFF
--- a/pkg/cmd/kitimport/kitfile_filter.go
+++ b/pkg/cmd/kitimport/kitfile_filter.go
@@ -1,0 +1,98 @@
+// Copyright 2025 The KitOps Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package kitimport
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/kitops-ml/kitops/pkg/artifact"
+	"github.com/kitops-ml/kitops/pkg/lib/filesystem/ignore"
+	repoutil "github.com/kitops-ml/kitops/pkg/lib/repo/util"
+)
+
+// kitfilePathFilter answers whether a given relative path would be packed
+// into some layer of the ModelKit defined by a Kitfile. It mirrors the layer
+// assignment that pack uses, so the HF download list agrees with what pack
+// will eventually include.
+type kitfilePathFilter struct {
+	ig     ignore.Paths
+	layers []string
+}
+
+func newKitfilePathFilter(kitfile *artifact.KitFile) (*kitfilePathFilter, error) {
+	ig, err := ignore.New(nil, kitfile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to process Kitfile to get file list: %w", err)
+	}
+	return &kitfilePathFilter{
+		ig:     ig,
+		layers: repoutil.LayerPathsFromKitfile(kitfile),
+	}, nil
+}
+
+// Matches returns true iff path would be packed into at least one layer.
+//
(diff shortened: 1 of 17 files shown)


pkg/output/sink.go
```diff
@@ -0,0 +1,146 @@
...
+//
+// Callers must ensure path is non-empty.
+func OpenDataSink(path string) (DataSink, error) {
+	if path == "-" {
+		sinkMu.Lock()
+		defer sinkMu.Unlock()
+		s := &stdoutSink{prevStdout: stdout}
+		stdout = stderr
+		return s, nil
+	}
```
amisevsk, line 120: Not an issue in the current usage, but what happens if you call `OpenDataSink` multiple times? If we open two sinks and then close them in the opposite order, does the original stdout get restored?

Maybe we need a guard to ensure only one stdout DataSink is open at a time?

pkg/cmd/kitimport/provenance.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// newProvenanceData constructs a ProvenanceData with the importer-invariant
+// fields filled in (start time, invocation ID, builder version).
+func newProvenanceData() *ProvenanceData {
+	return &ProvenanceData{
+		StartedOn:    time.Now().UTC(),
+		InvocationID: newInvocationID(),
+		KitVersion:   constants.Version,
+		GoVersion:    constants.GoVersion,
```
amisevsk, line 64: Should we also include the Kit commit for completeness? Version is technically mutable.

pkg/cmd/kitimport/kitfile_filter.go
```diff
@@ -0,0 +1,98 @@
...
+//
+// Both inputs are normalized to forward-slash form before comparison. Input
+// paths reach this filter as forward-slash (HF tree API; kfgen runs
+// filepath.ToSlash on filesystem walks), but layer paths go through
+// filepath.Clean in repoutil.LayerPathsFromKitfile, which on Windows rewrites
+// separators to backslashes for any layer with subdirectories. We do an
+// unconditional ReplaceAll rather than filepath.ToSlash so the comparison is
+// independent of GOOS — filepath.ToSlash is a no-op on non-Windows hosts and
+// would leave a Windows-cleaned layer string mismatched with a forward-slash
+// path during cross-platform handling.
```
amisevsk, line 87: Huh, today I learned. That's not clear at all from Golang's docs.

amisevsk (approved):

PR #1176, opened 2026-05-06
Init improvements

amisevsk:
### Description
Add a `--depth` flag for Kitfile generation (`kit init` and `kit import`) that can be used to control how far down a directory tree we traverse before starting to add less fine-grained directory entries. For `--depth=0` (the default), behaviour is the same as in the current main branch: all files in the context directory are classified and added to the Kitfile, and any subdirectories are added on a best-effort basis (if we can tell they're all one type, we use that type; otherwise we add a code layer for the directory).

Increasing the depth parameter means files in individual subdirectories are also classified in the same way, so `--depth=1` includes files in immediate subdirectories, `--depth=2` goes one directory further, etc. Depth can be set to `-1` to process all files individually. 

This flag can lead to much larger Kitfiles if a lot of individual files are added. However, this behaviour may be desirable in some cases, such as when a directory contains a large number of moderate or large files -- adding them as separate entries means they each appear in their own layer, and prevents the entire directory layer from being invalidated if a single file in the directory is updated.

Additionally, this PR makes one small change to how Kitfiles are generated: when directories are added to the Kitfile, we add a trailing slash to make it clearer that the layer represents a directory versus a regular file.

### Linked issues
N/A

### AI-Assisted Code

### DIFF
--- a/pkg/cmd/kitimport/cmd.go
+++ b/pkg/cmd/kitimport/cmd.go
@@ -80,6 +80,7 @@ type importOptions struct {
 	kitfilePath       string
 	downloadTool      string
 	concurrency       int
+	depth             int
 	attestationOutput string
 	modelKitRef       *registry.Reference
 }
@@ -102,6 +103,7 @@ func ImportCommand() *cobra.Command {
 	cmd.Flags().StringVarP(&opts.kitfilePath, "file", "f", "", "Path to Kitfile to use for packing (use '-' to read from standard input)")
 	cmd.Flags().StringVar(&opts.downloadTool, "tool", "", "Tool to use for downloading files: options are 'git' and 'hf' (default: detect based on repository)")
 	cmd.Flags().IntVar(&opts.concurrency, "concurrency", 5, "Maximum number of simultaneous downloads (for huggingface)")
+	cmd.Flags().IntVar(&opts.depth, "depth", 0, "Maximum directory depth to process when generating a Kitfile. Setting to -1 processes all files individually")
 	cmd.Flags().StringVar(&opts.attestationOutput, "attestation-output", "",
 		"Write SLSA Provenance v1 predicate to <path> ('-' for stdout)")
 	cmd.Flags().SortFlags = false
--- a/pkg/lib/kitfile/generate/generate.go
+++ b/pkg/lib/kitfile/generate/generate.go
@@ -263,13 +263,13 @@ func addDirToKitfile(kitfile *artifact.KitFile, dir DirectoryListing) (modelFile
 	case "docs":
 		output.Logf(output.LogLevelTrace, "Directory %s interpreted as documentation", dir.Name)
 		kitfile.Docs = append(kitfile.Docs, artifact.Docs{
-			Path: withTrailingSlash(dir.Path),
+			Path: unixWithTrailingSlash(dir.Path),
 		})
 		return nil
 	case "src", "pkg", "lib", "build":
 		output.Logf(output.LogLevelTrace, "Directory %s interpreted as code", dir.Name)
 		kitfile.Code = append(kitfile.Code, artifact.Code{
-			Path: withTrailingSlash(dir.Path),
+			Path: unixWithTrailingSlash(dir.Path),
 		})
 		return nil
 	}
@@ -309,7 +309,7 @@ func addDirToKitfile(kitfile *artifact.KitFile, dir DirectoryListing) (modelFile
 	}
 	if directoryHasMixedContents {
 		output.Logf(output.LogLevelTrace, "Mixed contents in directory %s, adding as code layer", dir.Path)
-		kitfile.Code = append(kitfile.Code, artifact.Code{Path: withTrailingSlash(dir.Path)})
+		kitfile.Code = append(kitfile.Code, artifact.Code{Path: unixWithTrailingSlash(dir.Path)})
 		return modelFiles
 	}
 	switch overallFiletype {
@@ -319,19 +319,19 @@ func addDirToKitfile(kitfile *artifact.KitFile, dir DirectoryListing) (modelFile
 		modelFiles = append(modelFiles, metadataFiles...)
 	case fileTypeDataset:
 		output.Logf(output.LogLevelTrace, "Interpreting directory %s as a dataset directory", dir.Path)
-		kitfile.DataSets = append(kitfile.DataSets, artifact.DataSet{Path: withTrailingSlash(dir.Path)})
+		kitfile.DataSets = append(kitfile.DataSets, artifact.DataSet{Path: unixWithTrailingSlash(dir.Path)})
 	case fileTypeDocs:
 		output.Logf(output.LogLevelTrace, "Interpreting directory %s as a docs directory", dir.Path)
-		kitfile.Docs = append(kitfile.Docs, artifact.Docs{Path: withTrailingSlash(dir.Path)})
+		kitfile.Docs = append(kitfile.Docs, artifact.Docs{Path: unixWithTrailingSlash(dir.Path)})
 	case fileTypePrompt:
 		output.Logf(output.LogLevelTrace, "Interpreting directory %s as a prompts directory", dir.Path)
-		kitfile.Prompts = append(kitfile.Prompts, artifact.Prompt{Path: withTrailingSlash(dir.Path)})
+		kitfile.Prompts = append(kitfile.Prompts, artifact.Prompt{Path: unixWithTrailingSlash(dir.Path)})
 	case fileTypeCode:
 		output.Logf(output.LogLevelTrace, "Interpreting directory %s as code directory", dir.Path)
-		kitfile.Code = append(kitfile.Code, artifact.Code{Path: withTrailingSlash(dir.Path)})
+		kitfile.Code = append(kitfile.Code, artifact.Code{Path: unixWithTrailingSlash(dir.Path)})
 	default:
 		output.Logf(output.LogLevelTrace, "Could not determine type for directory %s. Adding as code layer", dir.Path)
-		kitfile.Code = append(kitfile.Code, artifact.Code{Path: withTrailingSlash(dir.Path)})
+		kitfile.Code = append(kitfile.Code, artifact.Code{Path: unixWithTrailingSlash(dir.Path)})
 	}
 
 	return modelFiles
@@ -459,7 +459,8 @@ func anyPattern(query string, patterns []string) bool {
 	return false
 }
 
-func withTrailingSlash(p string) string {
+func unixWithTrailingSlash(p string) string {
+	p = filepath.ToSlash(p)
 	if p == "." || strings.HasSuffix(p, "/") {
 		return p
 	}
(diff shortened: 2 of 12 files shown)


pkg/lib/kitfile/generate/generate.go
```diff
@@ -173,98 +223,64 @@ func GenerateKitfile(dir *DirectoryListing, packageOpt *artifact.Package) (*arti
...
 
 	// Sort entries in the directory to try and figure out what it contains. We'll reuse the
 	// fact that the fileTypes are enumerated using iota (and so are ints) to index correctly.
 	// Avoid using maps here since they iterate in a random order.
 	directoryContents := [int(fileTypeUnknown) + 1][]string{}
 	for _, subdir := range dir.Subdirs {
-		// We can, in the future, recurse deeper into the directory tree here. For now, treat secondary dirs as unknowns
-		directoryContents[int(fileTypeUnknown)] = append(directoryContents[int(fileTypeUnknown)], subdir.Path)
+		// We can, in the future, recurse deeper into the directory tree here. For now, treat additional dirs as unknowns
+		directoryContents[int(fileTypeUnknown)] = append(directoryContents[int(fileTypeUnknown)], filepath.Join(withTrailingSlash(dir.Path), subdir.Path))
```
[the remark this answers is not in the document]
amisevsk: Ah yeah this was me getting confused because the test I had written was invalid :D

Forgot to remove this bit

pkg/lib/kitfile/generate/generate.go
```diff
@@ -113,15 +115,63 @@ func GenerateKitfile(dir *DirectoryListing, packageOpt *artifact.Package) (*arti
...
+	modelFiles, metadataFiles, unknownFiles, detectedLicenseType := processSubdirFiles(kitfile, *dir, depth)
+
+	if len(unknownFiles) > 5 {
+		output.Logf(output.LogLevelTrace, "Unrecognized files found in %s; adding catch-all code layer", dir.Path)
+		kitfile.Code = append(kitfile.Code, artifact.Code{Path: "."})
+	} else if len(unknownFiles) > 0 {
+		output.Logf(output.LogLevelTrace, "Unrecognized files found in %s; adding as code layers")
+		for _, f := range unknownFiles {
+			kitfile.Code = append(kitfile.Code, artifact.Code{Path: f.Path})
+		}
```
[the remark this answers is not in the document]
amisevsk: This is acceptable; replacing all detected code files with a catch-all layer is no longer desirable, in my opinon; we should avoid adding an opaque `Path: "."` layer if we can, as it's hard to reason about what it actually contains.

pkg/lib/kitfile/generate/generate.go
```diff
@@ -440,3 +458,10 @@ func anyPattern(query string, patterns []string) bool {
 	}
 	return false
 }
+
+func withTrailingSlash(p string) string {
+	if p == "." || strings.HasSuffix(p, "/") {
+		return p
+	}
+	return p + "/"
```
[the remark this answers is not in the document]
amisevsk: Paths should all be normalized during DirectoryListing generation (https://github.com/kitops-ml/kitops/blob/ff9a282403d339ca9ec7b64718aabb07835614eb/pkg/lib/kitfile/generate/filesystem.go#L79) so this should never see a windows-style backslash.

pkg/lib/kitfile/generate/generate.go
```diff
@@ -173,98 +223,64 @@ func GenerateKitfile(dir *DirectoryListing, packageOpt *artifact.Package) (*arti
 			kitfile.DataSets = append(kitfile.DataSets, artifact.DataSet{Path: file.Path})
 		case fileTypePrompt:
 			kitfile.Prompts = append(kitfile.Prompts, artifact.Prompt{Path: file.Path})
+		case fileTypeCode:
+			kitfile.Code = append(kitfile.Code, artifact.Code{Path: file.Path})
```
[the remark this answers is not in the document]
amisevsk: We don't currently check all the various source code file extensions but this approach is more correct and would have to be made if we started doing that. Not a bug to exhaust an enum in a switch.

pkg/lib/kitfile/generate/generate.go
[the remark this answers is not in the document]
gorkem: Is this enough? That normalization is inside the `t.IsRegular()` branch. It sets FileListing.Path, not `DirectoryListing.Path`.  Because of the way it works withTrailingSlash we may endup with `subdir-one\subdir-two/` on depth 2 or larger.

pkg/lib/kitfile/generate/generate.go
```diff
@@ -136,14 +186,14 @@ func GenerateKitfile(dir *DirectoryListing, packageOpt *artifact.Package) (*arti
...
 			})
 			continue
 		} else if strings.HasPrefix(strings.ToLower(file.Name), "license") {
 			output.Logf(output.LogLevelTrace, "Found license file '%s'", file.Name)
 			kitfile.Docs = append(kitfile.Docs, artifact.Docs{
-				Path:        file.Name,
+				Path:        file.Path,
 				Description: "License file",
 			})
 			licenseType, err := detectLicense(file.Path)
```
gorkem: we probably should stop license detection for subdirectories before wasting I/O since we discard them later in the code.

pkg/cmd/kitimport/cmd.go
```diff
@@ -100,6 +101,7 @@ func ImportCommand() *cobra.Command {
 	cmd.Flags().StringVarP(&opts.kitfilePath, "file", "f", "", "Path to Kitfile to use for packing (use '-' to read from standard input)")
 	cmd.Flags().StringVar(&opts.downloadTool, "tool", "", "Tool to use for downloading files: options are 'git' and 'hf' (default: detect based on repository)")
 	cmd.Flags().IntVar(&opts.concurrency, "concurrency", 5, "Maximum number of simultaneous downloads (for huggingface)")
+	cmd.Flags().IntVar(&opts.depth, "depth", 0, "Maximum directory depth to process when generating a Kitfile. Setting to -1 processes all files individually")
```
gorkem, line 106: Technically it is not just -1 any negative value.

pkg/lib/kitfile/generate/generate.go
```diff
@@ -113,15 +115,63 @@ func GenerateKitfile(dir *DirectoryListing, packageOpt *artifact.Package) (*arti
...
+	return kitfile, nil
+}
+
+// processSubdirFiles classifies each file in dir individually and adds typed layers to
+// kitfile. depth controls how many additional subdirectory levels are processed
+// file-by-file before switching to addDirToKitfile (whole-directory analysis). Returns
+// model and metadata files for the caller to pair, any files whose type could not be
+// determined (so the caller can decide how to handle them), and any detected license
+// identifier. Unclassifiable directories are added directly to kitfile.Code.
+func processSubdirFiles(kitfile *artifact.KitFile, dir DirectoryListing, depth int) (modelFiles []FileListing, metadataFiles []FileListing, unknownFiles []FileListing, detectedLicenseType string) {
```
gorkem: if this function classifies let's call it `classifyDirectoryRecursive` or something of that nature.

amisevsk: Done

pkg/cmd/kitimport/cmd.go
amisevsk: Updated the doc

pkg/lib/kitfile/generate/generate.go
amisevsk: Good point -- I rewrote the signature to return the path to the license file instead of the license and moved the detect step outside the recursive function.

pkg/lib/kitfile/generate/generate.go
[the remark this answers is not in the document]
amisevsk: Ah you're right. We're kind of messy in how paths are reconstructed from the DirectoryListing. I've updated the function to be `unixWithTrailingSlash` so that it also normalizes first.

pkg/lib/kitfile/generate/generate.go
```diff
@@ -113,15 +115,73 @@ func GenerateKitfile(dir *DirectoryListing, packageOpt *artifact.Package) (*arti
...
+			output.Debugf("Error determining license type: %s", err)
+			output.Logf(output.LogLevelWarn, "Unable to determine license type")
+		}
+		detectedLicenseType = licenseType
+		output.Logf(output.LogLevelTrace, "Detected license %s for license file", detectedLicenseType)
+	}
+
+	if len(unknownFiles) > 5 {
+		output.Logf(output.LogLevelTrace, "Unrecognized files found in %s; adding catch-all code layer", dir.Path)
+		kitfile.Code = append(kitfile.Code, artifact.Code{Path: "."})
```
gorkem, line 132: I think the appends is changing the catch-all behaviour. When I have a local folder with 6 root .sh files plus a src/ subdir. It produces `[{Path: "src/"}, {Path: "."}]` but I was expecting `[{Path: "."}].`

gorkem (approved):

pkg/lib/kitfile/generate/generate.go
amisevsk: Good catch -- this was a deliberate change as I found the behaviour of the catchall directory a little confusing (it is however something I will document with the next version's release notes).

Previously, the catch-all section was introduced as a fallback to ensure `kit init` always includes every file, and we figured that if we're adding a `code` layer that includes everything _else_ in the directory, it may as well be the only layer -- we replaced the entire `code` list with a single entry.

With the addition of the depth flag, this looks weirder -- I was testing the PR changes in the KitOps repo itself (despite that not being really the goal for KitOps), and was finding that it was really confusing as we'd detect documentation and yaml files, but they would _all_ intersect with the global `.` code layer. With the behaviour change we're discussing, any code layer (`src/` in your example) excludes that entire directory subtree from the catchall `.` layer.

I can revert the change so that at the default depth, we still get the single catch-all layer, but I still think the catch-all should be avoided as much as possible as it's entirely opaque -- there's no way to know what's in it without unpack it (whereas `src/` at least provides a hint).

Some example below the fold:

<details>
<summary>(expand)</summary>

```yaml
# Catch-all replaces all Code layers, depth=0 (default)
manifestVersion: 1.0.0
code:
  - path: .
---
# Catch-all replaces all Code layers, depth=-1 (full depth)
# You might be excused for thinking `src` contains only docs, and that the catch-all doesn't 
# include a `src` directory
manifestVersion: 1.0.0
code:
  - path: .
docs:
  - path: src/1-doc.md
  - path: src/2-doc.md
  - path: src/3-doc.md
  - path: src/4-doc.md
  - path: src/5-doc.md
---
# Current behaviour, depth=0 (default)
# This is the main behavioural change from before this PR; I personally think it's acceptable but
# hold that terribly strongly
manifestVersion: 1.0.0
code:
  - path: src/
  - path: .
---
# Current behaviour, depth=-1 (full depth)
manifestVersion: 1.0.0
code:
  - path: src/1.sh
  - path: src/2.sh
  - path: src/3.sh
  - path: src/4.sh
  - path: src/5.sh
  - path: .
docs:
  - path: src/1-doc.md
  - path: src/2-doc.md
  - path: src/3-doc.md
  - path: src/4-doc.md
  - path: src/5-doc.md
```

</details>

PR #1181, opened 2026-05-13
Add 10 more agents to the --as-skill registry

gorkem:
## Summary

Expands the list of coding agents supported by `kit unpack --as-skill`, so users of more agents can install SKILL.md prompt layers from a ModelKit directly into their agent's skills directory.

New agents:

- aider-desk
- codearts-agent
- codemaker
- codestudio
- devin
- dexto
- forgecode
- hermes-agent
- rovodev
- tabnine-cli

Each entry defines the project-relative skills directory, the global skills directory, and the detection probe used by `--as-skill` auto-discovery. Most fit the existing `homeAgent`/`homeAgentCustomGlobal` helpers; `devin` (XDG-based) and `tabnine-cli` (nested `.tabnine/agent/skills` global) are defined as literal entries.

Also replaces the prior single-agent spot-check test with a generic `TestAgentRegistryWellFormed` that iterates the entire registry. Adding future agents will not require updating the test.

## Test plan

- [x] `go build ./...`
- [x] `go fmt ./...` (no diffs after commit)
- [x] `go vet ./pkg/lib/skill/...`
- [x] `go test ./...` — full suite passes, including new registry test
- [x] `addlicense --check` on modified files
- [x] Manual smoke: `kit unpack <modelkit-with-SKILL.md> --as-skill=tabnine-cli -d /tmp/proj` and verify skill lands at `.tabnine/agent/skills/<name>`


amisevsk (approved):
No concerns here, though the Copilot comment does make sense re: the doc not matching

PR #1189, opened 2026-05-21
Respect KITOPS_HOME in library config path

### LINKED ISSUE #1171: Library usage of Kit does not take `KITOPS_HOME` into account
**Describe the bug**
A user can set `KITOPS_HOME` to override the default path for storage, credentials, etc. However, this environment variable is only processed in the root command and used to override the default value from constants. 

For usage that does not go through the root command, this means `KITOPS_HOME` is ignored.

The constants package should instead include a function that respects the `KITOPS_HOME` environment variable, with the same fallback logic as how the root command does it.

puneetdixit200:
### Description
Add `constants.ConfigPath()` so library callers can resolve the KitOps root storage directory with `$KITOPS_HOME` support and the existing platform default fallback. The root command now delegates environment/default path resolution to that helper while preserving `--config` flag precedence.

Regression coverage verifies both `$KITOPS_HOME` and default fallback behavior.

### Linked issues
Fixes #1171

### AI-Assisted Code

### Tests
- `GOMODCACHE=$PWD/.gomodcache GOCACHE=$PWD/.gocache go test ./pkg/lib/constants`
- `GOMODCACHE=$PWD/.gomodcache GOCACHE=$PWD/.gocache go test ./pkg/lib/constants ./cmd`
- `git diff --check`

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -193,13 +193,7 @@ func getConfigHome(opts *rootOptions) (string, error) {
 		return "", err
 	}
 
-	envHome := os.Getenv(constants.KitopsHomeEnvVar)
-	if envHome != "" {
-		output.Debugf("Using config directory from environment variable: %s", configHome)
-		return configHome, nil
-	}
-
-	output.Debugf("Using default config directory: %s", configHome)
+	output.Debugf("Using config directory: %s", configHome)
 	return configHome, nil
 }
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -188,20 +188,19 @@ func getConfigHome(opts *rootOptions) (string, error) {
 		return absHome, nil
 	}
 
-	envHome := os.Getenv(constants.KitopsHomeEnvVar)
-	if envHome != "" {
-		output.Debugf("Using config directory from environment variable: %s", envHome)
-		configHome, err := constants.ConfigPath()
-		if err != nil {
-			return "", err
-		}
-		return configHome, nil
-	}
-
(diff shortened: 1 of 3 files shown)


cmd/root.go
```diff
@@ -191,17 +191,17 @@ func getConfigHome(opts *rootOptions) (string, error) {
...
 
-	defaultHome, err := constants.DefaultConfigPath()
+	configHome, err := constants.ConfigPath()
 	if err != nil {
 		return "", err
 	}
-	output.Debugf("Using default config directory: %s", defaultHome)
-	return defaultHome, nil
+	output.Debugf("Using default config directory: %s", configHome)
+	return configHome, nil
```
amisevsk: Why is this doing the exact same thing, twice?

amisevsk (changes requested):

puneetdixit200:
Addressed in f401174. The root command now delegates the shared environment/default path resolution to `constants.ConfigPath()` and only applies the `--config` flag override separately, so the same resolution is not duplicated in `cmd/root.go`.

puneetdixit200:
The duplicate lookup concern is addressed in the current head: `getConfigHome` now applies the explicit `--config` override and then calls `constants.ConfigPath()` once for `KITOPS_HOME`/default resolution. I reran `go test ./pkg/lib/constants ./cmd` locally and it passes.

amisevsk (approved):
LGTM, thanks @puneetdixit200

PR #1214, opened 2026-06-18
feat: support packaging MCP Bundles (.mcpb) in ModelKits

gorkem:
## What

Adds first-class support for packaging MCP Bundles (`.mcpb`) in ModelKits via a new `mcpServers` Kitfile section.

- New `mcpServers` Kitfile section with the `MCPServer` entry type (name, path, description) and validation: required unique names, local-only paths, `.mcpb` extension, and cross-section duplicate-path checks.
- Each `.mcpb` bundle is validated as a well-formed ZIP containing a root `manifest.json`, then stored verbatim as its own **raw, uncompressed layer** (media type `application/vnd.kitops.modelkit.mcpb.v1.raw`). The layer digest covers the exact bundle bytes, so unpacking yields a byte-identical archive.
- `kit unpack` restores bundles to their Kitfile path; `--filter=mcpservers[:name|path]` selects them.
- `kit init` detects local `.mcpb` files (and directories of bundles) and adds them as `mcpServers` entries.
- `mcpServers` are rejected when packing in ModelPack format, since the media type is KitOps-local.

## Why

MCP Bundles are becoming a common way to distribute local MCP servers. Packaging them in ModelKits lets agents, models, and their MCP servers be versioned and shipped together as a single OCI artifact.

## Notes

Rather than introduce a bespoke verbatim-layer code path, this builds directly on the existing **raw-layer machinery** (`saveContentLayerAsRaw` / `extractRawLayer`): an `.mcpb` is simply a raw layer with `none` compression. The MCP-specific logic is limited to bundle validation, the `mcpServers` Kitfile schema, filtering, and `kit init` detection.

## Validation

- `go build ./...`, full `go test ./...` — all green
- Integration tests (`testing/mcpb_test.go`): digest == file SHA-256, byte-identical pack/unpack round-trip, `mcpservers` filter selection, and pack failures for malformed bundles and the ModelPack format
- Local CI mirror: `go mod tidy` (clean), `go fmt`, `addlicense` headers, `go generate` (no diff), trailing-whitespace check

### DIFF
--- a/pkg/lib/constants/mediatype/mediatype.go
+++ b/pkg/lib/constants/mediatype/mediatype.go
@@ -59,6 +59,8 @@ const (
 	CodeBaseType
 	// DocsBaseType is the base type for documentation layers
 	DocsBaseType
+	// MCPBBaseType is the base type for MCP Bundle (.mcpb) layers, stored as verbatim ZIP archives
+	MCPBBaseType
 )
 
 type CompressionType int
--- a/pkg/artifact/kitfile.go
+++ b/pkg/artifact/kitfile.go
@@ -51,13 +51,14 @@ const (
 
 type (
 	KitFile struct {
-		ManifestVersion string    `json:"manifestVersion" yaml:"manifestVersion"`
-		Package         Package   `json:"package,omitempty" yaml:"package,omitempty"`
-		Model           *Model    `json:"model,omitempty" yaml:"model,omitempty"`
-		Code            []Code    `json:"code,omitempty" yaml:"code,omitempty"`
-		DataSets        []DataSet `json:"datasets,omitempty" yaml:"datasets,omitempty"`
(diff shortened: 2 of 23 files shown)


pkg/lib/constants/mediatype/mediatype.go
```diff
@@ -59,6 +59,8 @@ const (
 	CodeBaseType
 	// DocsBaseType is the base type for documentation layers
 	DocsBaseType
+	// MCPBBaseType is the base type for MCP Bundle (.mcpb) layers, stored as verbatim ZIP archives
+	MCPBBaseType
```
amisevsk, line 63: I see that we catch invalid cases (CNCF ModelPack + MCP server) in other functions, but is there a way to move the error state into the mediatype package, e.g. in the ParseMediaType function? What gets output if you don't validate that the format is ModelKit?

amisevsk (approved):

PR #1222, opened 2026-06-23
fix: use %w instead of %s for error wrapping in fmt.Errorf calls

AdeshDeshmukh:
## Description

This PR fixes **11 instances** across **8 files** where `fmt.Errorf` was using `%s` to format an error argument instead of `%w`. While the error messages read just fine either way, using `%s` silently breaks the error chain — callers using `errors.Is()` or `errors.As()` cannot unwrap through these errors to inspect the underlying cause.

This is a Go best practice that the rest of the codebase already follows correctly (many adjacent lines use `%w` properly). These were likely copy-paste oversights.

### What was changed

Every fix is the same pattern: `%s` → `%w` for the error argument in a `fmt.Errorf` call.

| Package | File | Lines | Context |
|---------|------|-------|---------|
| `pkg/lib/filesystem/unpack` | `core.go`, `skill.go`, `util.go` | 4 | Wrapping `getStoreForRef`, `GetManifest`, and `NewLocalRepo` errors |
| `pkg/lib/repo/local` | `migration.go` | 2 | Wrapping `os.RemoveAll` errors during storage migration |
| `pkg/cmd/remove` | `remove.go` | 2 | Wrapping `removeModelRef` and `oras.Resolve` errors |
| `pkg/cmd/tag` | `tag.go` | 1 | Wrapping `oras.Resolve` error |
| `pkg/cmd/kitimport` | `util.go` | 1 | Wrapping `os.WriteFile` error |
| `pkg/cmd/dev` | `opts.go` | 1 | Wrapping `findAvailablePort` error |

### Additional cleanup

In `pkg/lib/filesystem/unpack/util.go`, the format string also had a stray trailing `\n` (`"failed to read local storage: %s\n"`), which was removed as part of the fix. Trailing newlines are unusual in Go error messages and can cause formatting issues in logs.

### Verification

- `go vet ./...` — ✅ passes
- `go build ./...` — ✅ compiles
- `go test ./...` — ✅ 63 tests pass across 6 affected packages
- DCO sign-off included

### Why this matters

Consider a caller trying to inspect the error from `removeModel()`:

```go
err := removeModel(ctx, opts)
if errors.Is(err, errdef.ErrNotFound) {
    // Handle not-found case
}
```

Before this fix, the `%s` verb stringifies the underlying error, so `errors.Is()` never matches — even though the codebase explicitly checks `err == errdef.ErrNotFound` in the same functions. After the fix, the error chain is preserved and unwrapping works as intended.

### Checklist

- [x] Changes are minimal and targeted — each fix is a single character change
- [x] `go vet` reports no issues
- [x] `go build` succeeds
- [x] Existing tests pass (63 tests across 6 packages)
- [x] DCO sign-off included

### DIFF
--- a/pkg/cmd/dev/opts.go
+++ b/pkg/cmd/dev/opts.go
@@ -106,7 +106,7 @@ func (opts *DevStartOptions) complete(ctx context.Context, args []string) error 
 	if opts.port == 0 {
 		availPort, err := findAvailablePort()
 		if err != nil {
-			return fmt.Errorf("invalid arguments: %s", err)
+			return fmt.Errorf("failed to find an available port: %w", err)
 		}
 		opts.port = availPort
 	}
--- a/pkg/cmd/kitimport/util.go
+++ b/pkg/cmd/kitimport/util.go
@@ -63,7 +63,7 @@ func generateKitfile(dirContents *kfgen.DirectoryListing, repo string, outDir st
 	}
 	kitfilePath := filepath.Join(outDir, constants.DefaultKitfileName)
 	if err := os.WriteFile(kitfilePath, kitfileBytes, 0644); err != nil {
-		return nil, fmt.Errorf("failed to write Kitfile: %s", err)
+		return nil, fmt.Errorf("failed to write Kitfile: %w", err)
 	}
 	output.Infof("Generated Kitfile:\n\n%s\n", string(kitfileBytes))
 	return kitfile, nil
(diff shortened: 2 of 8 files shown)


AdeshDeshmukh:
Agreed that message was misleading. Updated in the latest push.

amisevsk (approved):
Looks good to me, thanks @AdeshDeshmukh

PR #1225, opened 2026-06-24
feat: respect SOURCE_DATE_EPOCH for reproducible manifest digests

### LINKED ISSUE #1224: support SOURCE_DATE_EPOCH for reproducible pack digests

PR #1139 added the org.opencontainers.image.created OCI annotation to ModelKit manifests — but hardcoded time.Now() as the value. This makes it impossible to produce stable, reproducible digests across runs of the same content.

Respect the `SOURCE_DATE_EPOCH` environment variable, idiomatic for pinning build timestamps in [Docker BuildKit](https://github.com/moby/buildkit/blob/master/docs/build-repro.md). When set, use its Unix timestamp value for the created annotation. When unset, keep the current behavior. When set to a non-integer, fail with a clear error rather than silently falling back to wall-clock time (matching BuildKit's behavior).

A CLI flag (`--source-date-epoch`) was considered, but the environment variable is the established convention and is automatically propagated by Nix dev shells, BuildKit, and CI reproducibility pipelines without any user configuration.

ericsigurdson-dial:
**Description**

Addresses the concern raised in #1139's review: the `org.opencontainers.image.created` annotation added in #1139 is always `time.Now()`, making reproducible digests impossible for the same content packed at different times.

This change makes createManifest in `pkg/lib/filesystem/local-storage.go` respect `SOURCE_DATE_EPOCH`, standard used by Docker BuildKit and reproducible-build toolchains.

 Behavior:
  - `SOURCE_DATE_EPOCH` unset → `time.Now()` (unchanged)
  - `SOURCE_DATE_EPOCH=<unix seconds>` → manifest created uses that timestamp; digest is stable across runs
  - `SOURCE_DATE_EPOCH=<non-integer>` → pack fails with a clear error, matching [BuildKit's behavior](https://github.com/moby/buildkit/blob/master/docs/build-repro.md)

 Verified:
```
$ SOURCE_DATE_EPOCH=abc ./kit pack -t foo:v1 .
[ERROR] Failed to pack model kit: error creating manifest: invalid SOURCE_DATE_EPOCH "abc": ...
```
```
$ SOURCE_DATE_EPOCH=315532800 ./kit pack -t foo:v1 .
Model saved: sha256:cd204...   # stable across runs with same content
```
**Linked issues**

closes #1224

**AI-Assisted Code**

### DIFF
--- a/pkg/lib/filesystem/local-storage.go
+++ b/pkg/lib/filesystem/local-storage.go
@@ -21,6 +21,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
+	"strconv"
 	"time"
 
 	"github.com/kitops-ml/kitops/pkg/artifact"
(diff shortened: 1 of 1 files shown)


gorkem:
please fix dco

ericsigurdson-dial:
> please fix dco

fixed!

amisevsk (approved):
Thanks @ericsigurdson-dial! I like reusing the docker standard here.

PR #1230, opened 2026-06-25
Fixup duplicated default documented in pack command

amisevsk:
### Description
The Cobra library automatically documents default values, so listing the default in flag descriptions is not necessary.

This PR does update the cli-reference.md doc outside of a release, but the change is minor enough (wording, not correctness) that I don't think a bugfix release is necessary here.

### AI-Assisted Code


gorkem (approved):
LGTM

PR #1231, opened 2026-06-27
fix: remove redundant double-close in HuggingFace download

AdeshDeshmukh:
## Description

Noticed a `TEMP:` marker in `pkg/lib/external/hf/download.go` — this resolves it.

The outer `defer resp.Body.Close()` was redundant: `contentRC.Close()` (from `progress.TrackDownload`) always reaches `resp.Body.Close()` — whether directly (progress disabled) or through mpb's `proxyReader` which delegates `Close()` to the embedded `io.ReadCloser`. This caused a double-close on every successful download.

### Changes

- **Removed** the outer `resp.Body.Close()` defer (redundant, caused double-close)
- **Added** explicit `resp.Body.Close()` before the early return on non-200 status (previously handled by the now-removed defer)
- **Fixed** the `TEMP:` message with a descriptive log including the filename

### Risk

Low. `io.Copy` fully drains the body before close runs, so HTTP connection reuse is preserved. All existing tests pass (13/13). `go vet` clean.

### AI-Assisted Code

### DIFF
--- a/pkg/lib/external/hf/download.go
+++ b/pkg/lib/external/hf/download.go
@@ -127,20 +127,17 @@ func downloadFile(
 	if err != nil {
 		return fmt.Errorf("error calling API: %w", err)
 	}
-	defer func() {
-		if err := resp.Body.Close(); err != nil {
-			plog.Logf(output.LogLevelWarn, "Failed to close response body: %s", err)
-		}
-	}()
-
 	if resp.StatusCode != http.StatusOK {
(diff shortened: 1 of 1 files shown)


AdeshDeshmukh:
@javisperez sir. PTAL

amisevsk (approved):
Looks good to me, thanks @AdeshDeshmukh. I merged the copilot suggestion to log the error as before, which broke DCO, so I squashed it into your commit and force pushed.

PR #1243, opened 2026-07-09
feat: support packing and pushing directly to a remote registry

### LINKED ISSUE #1026: Enable packing and pushing directly to remote without storing layers locally
While packing a ModelKit using `kit pack`, Kit will

1. Determine which files need to be included in a layer
2. Pack the files into a ModelKit layer
3. Calculate the layer's digest
4. Add it to the local cache

After a ModelKit is packed, a common next step is to push it to a remote, after which the local copy can be deleted to save space. The local cache is currently necessary as we need to compute the digests of all layers in order to compute the digest of the ModelKit manifest.

However, storing files locally introduces some inefficiency, especially in CI environments where disk storage is a limited resource. Each ModelKit effectively must be stored twice -- once in the local directory being packed, and once it the local Kit cache -- before being pushed to a remote. As Models are often very large files, this inefficiency can cause storage requirements to become very high.

Instead, we should support packing a ModelKit directly to a remote (e.g. using a flag on `kit pack`):

* For each layer, initiate an upload with a remote registry and stream the layer directly into the remote registry while calculating the digest of the pushed data, skipping local storage
* Once all layers have been pushed, generate a manifest from the calculated layer manifests and push that to the remote

This is possible because layer uploads in OCI only require the file digest upon completion, meaning we don't need to know the digest prior to push.

**Additional context**
N/A

Hemanth-1354:
### Description
Adds a `--push` flag to `kit pack` that streams the packed modelkit directly to
the remote registry specified by `--tag`, skipping local storage entirely.

Currently `kit pack` always writes layers/config/manifest to the local on-disk
cache before an optional separate `kit push`, meaning a model is stored twice
(source directory + local cache) before ever reaching the registry — costly in
CI environments where disk is limited. With `--push`, `SaveModel` and its layer
helpers write to a generic `oras.Target` instead of a concrete local repo, so
the same pack pipeline can target either local storage (default, unchanged
behavior) or a remote registry (`remote.NewRepository`) directly. Digests are
still computed via a temporary file (required to know the digest before
pushing), but nothing is persisted to the permanent local ModelKit cache when
`--push` is used.

Follows the approach agreed with @amisevsk in the issue thread (flag named
`--push` per that discussion).

### Linked issues
Closes #1026

### AI-Assisted Code

### DIFF
--- a/docs/src/docs/cli/cli-reference.md
+++ b/docs/src/docs/cli/cli-reference.md
@@ -750,9 +750,6 @@ at the root of the provided context directory. Any relative paths defined
 within the kitfile are interpreted as being relative to this context
 directory.
 
-If --push is set along with a --tag pointing to a remote registry, the modelkit
-is streamed directly to that registry instead of being stored locally.
-
 ```
 kit pack [flags] DIRECTORY
 ```
@@ -765,9 +762,6 @@ kit pack .
 
 # Pack a modelkit with a specific kitfile and tag
 kit pack . -f /path/to/your/Kitfile -t registry/repository:modelv1
-
-# Pack a modelkit and push it directly to a remote registry without storing it locally
-kit pack . -t registry.example.com/my-org/my-model:latest --push
 ```
 
 ### Options
@@ -778,14 +772,6 @@ kit pack . -t registry.example.com/my-org/my-model:latest --push
       --compression string    Compression format to use for layers. Valid options: 'none', 'gzip', 'gzip-fastest', 'zstd' (default "none")
       --layer-format string   Packaging format to use for layers. Valid options: 'tar', 'raw' (default "tar")
       --use-model-pack        Pack model in ModelPack format instead of ModelKit
-      --push                  Stream the packed modelkit directly to the remote registry specified by --tag, without storing it locally
-      --plain-http            Use plain HTTP when connecting to remote registries
-      --tls-verify            Require TLS and verify certificates when connecting to remote registries (default true)
-      --tls-cert strings      Path to TLS cert to add to trust store (flag can be repeated)
-      --cert string           Path to client certificate used for authentication (can also be set via environment variable KITOPS_CLIENT_CERT)
-      --key string            Path to client certificate key used for authentication (can also be set via environment variable KITOPS_CLIENT_KEY)
-      --concurrency int       Maximum number of simultaneous uploads/downloads (default 5)
-      --proxy string          Proxy to use for connections (overrides proxy set by environment)
   -h, --help                  help for pack
 ```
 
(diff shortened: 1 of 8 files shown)


docs/src/docs/cli/cli-reference.md
amisevsk, line 1: Please revert changes to cli-reference.md -- this file is automatically regenerated from CLI help pages when a new release is generated.

Otherwise, this documentation appears on https://kitops.org/docs/cli/cli-reference/ before the flags are actually available in a release.

pkg/cmd/pack/cmd.go
```diff
@@ -155,6 +166,15 @@ func (opts *packOptions) complete(ctx context.Context, args []string) error {
...
 	}
 
+	if opts.push {
+		if opts.modelRef.Registry == artifact.DefaultRegistry {
+			return fmt.Errorf("--push requires a remote registry to be specified with --tag")
+		}
+		if err := opts.NetworkOptions.Complete(ctx, args); err != nil {
+			return err
+		}
+	}
```
amisevsk, line 178: If push is false, we should print a warning that network flags are only applicable when `--push` is specified.

amisevsk (changes requested):
Thanks @Hemanth-1354. I have a couple of minor comments, but otherwise this looks generally good. If you drop the changes to `cli-reference.md` and add the warning log on flags, this looks good to merge.

This doesn't fully hit what I was thinking for the issue, but is a great improvement nonetheless. A future (and significantly more complicated) improvement would be to do the push _without_ storing the intermediate file on disk -- the tar writer could write directly into a remote connection and send the digest to seal the upload.

pkg/cmd/pack/cmd.go
Hemanth-1354: Done. added warning at pkg/cmd/pack/cmd.go:177. Uses cmd.Flags().Changed() so it only triggers when user explicitly passes network flags without --push.

docs/src/docs/cli/cli-reference.md
Hemanth-1354: Done. reverted docs/src/docs/cli/cli-reference.md to main state. Zero diff now. The flags are documented in the Cobra help text (--push, --plain-http, etc.) which the auto-generator will pick up at release time.

Hemanth-1354:
﻿Addressed all review feedback.

Thanks for the review @amisevsk.

1. Reverted `docs/src/docs/cli/cli-reference.md`, understood it's auto-generated at release time. Flags are already documented in Cobra help text so the generator picks them up.

2. Added warning in `pkg/cmd/pack/cmd.go:177`, when network flags (`--plain-http`, `--tls-verify`, `--tls-cert`, `--cert`, `--key`, `--concurrency`, `--proxy`) are explicitly passed without `--push`, a warning prints: "Warning: network-related flags are only applicable with --push flag". Uses `cmd.Flags().Changed()` so defaults don't trigger it.

PR #1247, opened 2026-07-11
fix: return error on extra tag removal failure

### LINKED ISSUE #1246: kit remove exits 0 when extra tags fail to remove
**Describe the bug**

`kit remove` exits with code 0 even when extra comma-separated tags fail to remove. The function `removeModel()` logs the errors via `output.Errorf` but always returns `nil`, so the caller never knows a failure happened. This means `kit remove myrepo:tag1,tag2,bogus` silently reports success despite bogus failing.

**To Reproduce**

```
kit pull alice/my-model:latest
kit tag alice/my-model:latest alice/my-model:extra1 alice/my-model:extra2
kit remove alice/my-model:latest,extra1,extra2,nonexistent
echo $?
```

`extra2` and `nonexistent` both fail (they don't exist), but exit code is `0`.

**Version**

latest

**Additional context**

The loop at `pkg/cmd/remove/remove.go:115-126` iterates `opts.extraTags`, logs failures, but returns `nil`. Compare with lines 108-110 where the primary reference correctly returns an error.

Workaround: remove each tag individually instead of comma-separated.

AdeshDeshmukh:
### Description
removeModel() in pkg/cmd/remove/remove.go logs errors when extra
comma-separated tags fail to remove but always returns nil. This
causes `kit remove repo:tag1,tag2,bogus` to exit with code 0 even
when some tags fail, since the caller only checks for a non-nil error.

Changes:
- Track failure count in the extra tags loop
- Return a non-nil error if any extra tag removal failed
- Fix a pre-existing %s→%w format verb in the same function

The exit code now correctly reflects partial failures.

### Linked issues
Closes #1246

### AI-Assisted Code

### DIFF
--- a/pkg/cmd/remove/remove.go
+++ b/pkg/cmd/remove/remove.go
@@ -107,11 +107,12 @@ func removeModel(ctx context.Context, opts *removeOptions) error {
 	}
 	desc, err := removeModelRef(ctx, localRepo, opts.modelRef, opts.forceDelete)
 	if err != nil {
-		return fmt.Errorf("failed to remove: %s", err)
+		return fmt.Errorf("failed to remove: %w", err)
 	}
 	displayRef := artifact.FormatRepositoryForDisplay(opts.modelRef.String())
 	output.Infof("Removed %s (digest %s)", displayRef, desc.Digest)
 
+	var failedCount int
 	for _, tag := range opts.extraTags {
 		ref := *opts.modelRef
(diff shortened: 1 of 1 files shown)


AdeshDeshmukh:
@javisperez sir. PTAL

amisevsk (approved):
Looks reasonable to me, thanks @AdeshDeshmukh

Issue #1249, opened 2026-07-18
Add a kit skill command to teach AI agents how to use the kit CLI


gorkem:
## Problem

Coding agents (Claude Code, Codex, Cursor, Copilot, Gemini CLI) increasingly drive CLIs directly. Today an agent's only way to learn kit is `--help` output, which is flag-oriented and says nothing about workflows: Kitfile authoring, pack/push/pull/unpack, tagging. Users who want agents working with ModelKits hand-write AGENTS.md sections or paste docs into context, and those drift from the installed kit version.

Other agent-facing CLIs have converged on shipping this knowledge as an installable skill following the [Agent Skills spec](https://agentskills.io):

- **hunk** ships `skills/hunk-review/SKILL.md` in-repo and provides a `hunk skill` command to locate/load it
- **beads** has `bd setup <agent>` to install a skill and agent instructions, plus `bd prime` to print workflow context on demand

## Proposal

Add a `kit skill` command that installs a SKILL.md (covering Kitfile authoring, the pack/push/pull/unpack workflow, tagging, and common errors) into the appropriate agent directory (`.claude/skills/`, the shared `.agents/skills/`, etc.).

The skill should be embedded in the binary so it always matches the installed kit version and works air-gapped.

We should also publish the skill as a ModelKit. It's the natural distribution channel for our own project: the skill becomes a versioned, verifiable OCI artifact that agents and CI can pull like any other ModelKit, and it demonstrates skills-as-artifacts with our own tooling.

## Related

`kit unpack --as-skill`: unpacking a ModelKit directly into an agent skills directory. The kit CLI skill published as a ModelKit would be a natural first consumer of that path.

PR #1252, opened 2026-07-21
feat(skill): add a kit skill command to install the kit agent skill

youdie006:
### Description

Coding agents (Claude Code, Codex, Cursor, ...) learn `kit` only from flag-oriented `--help`, which says nothing about workflows. This adds a `kit skill` command that installs an embedded `SKILL.md` — covering Kitfile authoring and the pack/tag/login/push/pull/unpack workflow, plus common errors — into the agent skills directories.

- New `pkg/cmd/skill` command, registered in `cmd/root.go` alongside the other subcommands.
- The `SKILL.md` is `go:embed`'d so it always matches the installed kit version and works air-gapped, exactly as the issue asks.
- It reuses the existing `--as-skill` installer (`libskill.InstallSkill`) and agent-directory resolution, so it lands in `.claude/skills`, the shared `.agents/skills`, etc., with the same `--agents` / `-d` / `-o` / `-i` flags as `kit unpack --as-skill`. No agent-dir logic is duplicated.
- Every section of the embedded `SKILL.md` is grounded in the real command implementations and Kitfile validation rules (no invented flags/behavior).
- `docs/src/docs/cli/cli-reference.md` regenerated via the repo's own `docs/src/docs/cli/generate.sh`.

Scoped to the core command per the small-PR principle. The issue's secondary ask — publishing the skill as a ModelKit — is left to a follow-up.

Tests (`pkg/cmd/skill/cmd_test.go`): installs the embedded skill to `.claude/skills/kitops/SKILL.md`; multi-agent install; unknown-agent error; a frontmatter drift guard. `go build ./...`, `go vet`, `go test`, and `gofmt` all pass; also smoke-tested with the built binary (writes the skill, overwrite / ignore-existing guards work, `--help` renders).

Design note: I named the skill/dir `kitops`; if you'd prefer something more action-oriented (e.g. `using-kit`) it's a one-line change to the `skillName` const + the frontmatter `name` (a test guards they stay equal).

### Linked issues

Refs #1249

### AI-Assisted Code


docs/src/docs/cli/cli-reference.md
amisevsk, line 1: A bit of a strange feature in this repository: this CLI reference doc is served at https://kitops.org/docs/cli/cli-reference/ based on the repository HEAD (i.e. not the last release). As a result, merging this update to the `cli-reference.md` file will update the public docs immediately, even though the `skill` command is not yet released.

Please revert this change in the PR; instead, this file will be automatically updated on the next Kit release through CI actions.

pkg/cmd/skill/SKILL.md
amisevsk, line 1: @gorkem I'd like your opinion on this SKILL.md for KitOps -- you're better at writing these than I am.

amisevsk (changes requested):
Nicely done @youdie006, thank you!

I have one note on the `cli-reference.md` file, but apart from that it looks good to me. I'd like @gorkem's review on the SKILL.md, though I have no concerns personally and think it's a solid starting point.

youdie006:
Thanks @amisevsk! Good catch on the CLI reference being served from HEAD - I've dropped the regenerated `cli-reference.md` from this PR so merging won't surface `kit skill` in the live docs before a release; the reference will pick it up when it's regenerated at release time. The command, its embedded SKILL.md, and the root registration remain. Left the SKILL.md for @gorkem's review as you suggested.

gorkem (approved):

youdie006:
@amisevsk gentle nudge on this one — the `cli-reference.md` regen is dropped (05a9a24), so merging won't surface `kit skill` in the live docs before a release, and @gorkem has approved the SKILL.md. Whenever you have a moment, could you re-review so we can unblock it?

amisevsk (approved):
Sorry for the delay, I got busy with other things.

Thanks @youdie006!

PR #1260, opened 2026-08-05
Update the download links in the docs

javisperez:
This PR updates the download links for binaries to go through Scarf rather than directly to github.


Jwilliamsr (approved):
Thanks Javis

